PR #21, opened 2015-01-07
Allow transformers 0.4

snoyberg:


pbrisbin:
Thanks! Will release with this change today.

Issue #23, opened 2015-03-16
Support `state` to prevent CSRF


jprider63:
Hi,

I was wondering if you'd be interesting in adding support for the `state` field to prevent CSRF attacks. I think you'd just need to modify the callback dispatch like `oauth2GithubScoped` does. 

JP

pbrisbin:
I don't know a ton about this. Generally, if adding something makes things safer and doesn't require anything from existing or new end-users of the library, then I don't see why not.

jprider63:
I took a shot at this here: d33cc46b87a9efce5421cc076b1b243363fe8b72

The main compatibility issue I see is when plugins manually set the state (like Github). I think it should be relatively painless to change these though.

pbrisbin:
The diff looks reasonable enough to me. Would you like to move it to a proper PR and pickup discussion/review there?

PR #24, opened 2015-03-18
Add `state` to prevent CSRF

jprider63:
Added `state` field to prevent CSRF. Removed Github's state implementation it should be subsumed. 

Previous discussion at #23.

### DIFF
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -14,11 +14,12 @@ module Yesod.Auth.OAuth2
     , module Network.OAuth.OAuth2
     ) where
 
+import Control.Applicative ((<$>))
 import Control.Exception.Lifted
 import Control.Monad.IO.Class
 import Data.ByteString (ByteString)
-import qualified Data.ByteString as BS
-import Data.Text (Text, append, pack)
+import Data.Monoid ((<>))
+import Data.Text (Text, pack)
 import Data.Text.Encoding (decodeUtf8With, encodeUtf8)
 import Data.Text.Encoding.Error (lenientDecode)
 import Data.Typeable
@@ -57,7 +58,7 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
         withCallback csrfToken = do
             tm <- getRouteToParent
             render <- lift $ getUrlRender
-            let newEndpoint = oauthOAuthorizeEndpoint oauth `BS.append` "&state=" `BS.append` encodeUtf8 csrfToken
+            let newEndpoint = oauthOAuthorizeEndpoint oauth <> "&state=" <> encodeUtf8 csrfToken
             return $ oauth { 
                     oauthCallback = Just $ encodeUtf8 $ render $ tm url, 
                     oauthOAuthorizeEndpoint = newEndpoint
@@ -66,7 +67,7 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
         dispatch "GET" ["forward"] = do
             csrfToken <- liftIO $ generateToken
             setSession tokenSessionKey csrfToken
-            authUrl <- fmap (bsToText . authorizationUrl) $ withCallback csrfToken
+            authUrl <- (bsToText . authorizationUrl) <$> withCallback csrfToken
             lift $ redirect authUrl
 
         dispatch "GET" ["callback"] = do
@@ -89,10 +90,10 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
 
         dispatch _ _ = notFound
 
-        generateToken = fmap (pack . take 30 . randomRs ('a','z')) newStdGen
+        generateToken = (pack . take 30 . randomRs ('a','z')) <$> newStdGen
 
         tokenSessionKey :: Text
-        tokenSessionKey = "_yesod_oauth2_" `append` name
+        tokenSessionKey = "_yesod_oauth2_" <> name
 
         login tm = do
             render <- getUrlRender
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -17,12 +17,14 @@ module Yesod.Auth.OAuth2
 import Control.Exception.Lifted
 import Control.Monad.IO.Class
 import Data.ByteString (ByteString)
-import Data.Text (Text)
+import qualified Data.ByteString as BS
+import Data.Text (Text, append, pack)
 import Data.Text.Encoding (decodeUtf8With, encodeUtf8)
 import Data.Text.Encoding.Error (lenientDecode)
 import Data.Typeable
 import Network.OAuth.OAuth2
 import Network.HTTP.Conduit(Manager)
+import System.Random
 import Yesod.Auth
 import Yesod.Core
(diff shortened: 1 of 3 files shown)


Yesod/Auth/OAuth2.hs
```diff
@@ -52,28 +54,46 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
     where
         url = PluginR name ["callback"]
 
-        withCallback = do
+        withCallback csrfToken = do
             tm <- getRouteToParent
             render <- lift $ getUrlRender
-            return $ oauth { oauthCallback = Just $ encodeUtf8 $ render $ tm url }
+            let newEndpoint = oauthOAuthorizeEndpoint oauth `BS.append` "&state=" `BS.append` encodeUtf8 csrfToken
```
pbrisbin: Is this state parameter a part of the OAuth2 spec? I.e. can we be reasonable sure all plugins built on top of this will work?

Yesod/Auth/OAuth2/Github.hs
```diff
@@ -77,27 +77,27 @@ oauth2GithubScoped clientId clientSecret scopes = basicPlugin {apDispatch = disp
...
+        -- dispatch "GET" ["callback"] = do
+        --     state <- lift $ runInputGet $ ireq textField "state"
+        --     savedState <- lookupSession "githubState"
+        --     _ <- apDispatch basicPlugin "GET" ["callback"]
+        --     case savedState of
+        --         Just saved | saved == state -> apDispatch basicPlugin "GET" ["callback"]
+        --         Just saved -> invalidArgs ["state: " `mappend` state `mappend` ", and not: " `mappend` saved]
+        --         _ -> invalidArgs ["state: " `mappend` state]
+
+        -- dispatch method ps = apDispatch basicPlugin method ps
```
pbrisbin: This is awesome to see. Please go ahead and delete it, not need to keep commented code around.

Yesod/Auth/OAuth2/Github.hs
```diff
@@ -77,27 +77,27 @@ oauth2GithubScoped clientId clientSecret scopes = basicPlugin {apDispatch = disp
...
                 }
 
-        withState state = authOAuth2 "github"
-            (oauth {oauthOAuthorizeEndpoint = oauthOAuthorizeEndpoint oauth `BS.append` "&state=" `BS.append` encodeUtf8 state})
-            fetchGithubProfile
+        -- withState state = authOAuth2 "github"
+        --     (oauth {oauthOAuthorizeEndpoint = oauthOAuthorizeEndpoint oauth `BS.append` "&state=" `BS.append` encodeUtf8 state})
+        --     fetchGithubProfile
 
         basicPlugin = authOAuth2 "github" oauth fetchGithubProfile
```
pbrisbin: I would just inline this.

yesod-auth-oauth2.cabal
```diff
@@ -36,6 +36,7 @@ library
                    , aeson                   >= 0.6       && < 0.9
                    , yesod-core              >= 1.2       && < 1.5
                    , authenticate            >= 1.3.2.7   && < 1.4
+                   , random
```
pbrisbin: Is there a reason to add this and do what you're doing vs just moving what the GitHub plugin was doing with `UUID.nextRandom` into the base plugin? If so, can we remove a `uuid` dependency here too?

Yesod/Auth/OAuth2.hs
jprider63: The state parameter is recommended according to section 4.1.1 in the [RFC](http://tools.ietf.org/html/rfc6749). If you're worried about sites that don't support it, we could also add a `authOAuth2NoState` function for compatibility?

pbrisbin: I think if we can confirm all the services here (GitHub, Learn, Spotify, and sort of Google) support the state parameter, then let's not worry about it. We know GitHub does, I can look into Learn. Could you look into Spotify and Google?

jprider63: Ok makes sense. Spotify and Google both support it.

yesod-auth-oauth2.cabal
jprider63: No real reason. I'm not too familiar with UUIDs and any benefits they may provide over a random token. I'm fine with removing the `uuid` dependency.

Yesod/Auth/OAuth2/Github.hs
```diff
@@ -18,16 +18,16 @@ import Control.Exception.Lifted
...
-import Yesod.Form
+-- import Yesod.Core
+-- import Yesod.Form
 import Network.HTTP.Conduit(Manager)
-import Data.UUID (toString)
-import Data.UUID.V4 (nextRandom)
-import qualified Data.ByteString as BS
+-- import Data.UUID (toString)
+-- import Data.UUID.V4 (nextRandom)
+-- import qualified Data.ByteString as BS
```
pbrisbin: Some more dead code.

Yesod/Auth/OAuth2.hs
```diff
@@ -17,12 +17,14 @@ module Yesod.Auth.OAuth2
 import Control.Exception.Lifted
 import Control.Monad.IO.Class
 import Data.ByteString (ByteString)
-import Data.Text (Text)
+import qualified Data.ByteString as BS
+import Data.Text (Text, append, pack)
```
pbrisbin: Would you mind importing `Data.Monoid.(<>)` and using that consistently in place of `Text.append`, `BS.append`?

Yesod/Auth/OAuth2.hs
pbrisbin: To confirm, I think this is fine.

Yesod/Auth/OAuth2.hs
```diff
@@ -52,28 +54,46 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
...
+            return $ oauth { 
+                    oauthCallback = Just $ encodeUtf8 $ render $ tm url, 
+                    oauthOAuthorizeEndpoint = newEndpoint
+                }
 
         dispatch "GET" ["forward"] = do
-            authUrl <- fmap (bsToText . authorizationUrl) withCallback
+            csrfToken <- liftIO $ generateToken
+            setSession tokenSessionKey csrfToken
+            authUrl <- fmap (bsToText . authorizationUrl) $ withCallback csrfToken
```
pbrisbin: Can you switch any calls like `fmap f $ g x` to `f <$> g x`?

pbrisbin:
Some more minor comments. I'm in favor of this change.

jprider63:
Cool. I just addressed your latest comments. Let me know if you see anything else.

pbrisbin:
> Let me know if you see anything else.

Will do.

Assuming I don't, do you mind if I rebase things into one commit before merging (you'll still retain credit, I might just modify the message a bit)? I should have time to do this after work tonight.

jprider63:
That's fine with me. I can give you write access to the branch if that'd make it easier for you.

pbrisbin:
No need, but thanks anyway

pbrisbin:
Merged as bc320b1

PR #27, opened 2015-04-03
Add Twitter as provider

pbrisbin:
- Extract generic OAuth2Plugin construct
- Define Twitter in terms of it
- Will re-implement Upcase, GitHub, etc in terms of it separately


pbrisbin:
Ping @thoughtbot/haskell 

I'm actually thinking we won't use this in Carnival, since Twitter won't give us an email. That said, I think it's still valuable and I'd like to merge and follow up with the refactor I mentioned in the PR description.

pbrisbin:
I'm going to open a new PR for this.

PR #28, opened 2015-04-07
Update for GHC 7.10 and add a travis.yml

pbrisbin:


.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: haskell
+
+ghc:
```
jferris: Why do we install all these versions?

pbrisbin: I just put everything for now, see what works. No one has ever said what we do and don't support.

jferris: Ah, okay. I think we should only support the version we plan to deploy to.

pbrisbin: Deploy what, Carnival? This is a general purpose library, I think we should support as much as we can, to a reasonable limit. I think latest and 1 or 2 back is reasonable, assuming supporting 7.6 isn't a huge jungle of CPP. Since the decision's not been thought about yet, I think it makes sense to see what it currently works with now and support that until something prompts us to think about it again.

jferris: Ha, I thought I was commenting on Carnival. Yeah, I think what you're doing makes sense here.

pbrisbin:
Updated to do a few fixes, kept them as isolated commits.

PR #30, opened 2015-04-13
Extract common functionality from Plugins

pbrisbin:
This intended to extract common functionality into an OAuth2Plugin data type.
It came about when adding Twitter in #27. I notice now from the diff-stat that
it seems to not have removed much duplication. Would appreciate other's
opinions.

This does include some behavior changes

- GitHub (scoped version) and Spotify now accept the scopes argument first,
  otherwise we'd have to clutter their definition with named arguments for the
  client id/secret that are passed along to the generic constructor
- Spotify's scopes argument is now `[Text]`, not `[ByteString]`, this makes it
  consistent with the rest of the library


pbrisbin:
I'm going to do this more simply ass discussed

PR #31, opened 2015-04-13
Add Twitter as OAuth provider

pbrisbin:
- Extract fromProfileURL and reuse

### DIFF
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -12,6 +12,7 @@
 module Yesod.Auth.OAuth2
     ( authOAuth2
     , oauth2Url
+    , fromProfileURL
     , YesodOAuth2Exception(..)
     , module Network.OAuth.OAuth2
     ) where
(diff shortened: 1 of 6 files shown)


pbrisbin:
Ping @thoughtbot/haskell

Yesod/Auth/OAuth2.hs
```diff
@@ -104,5 +108,21 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
         <a href=@{tm $ oauth2Url name}>Login via #{name}
         |]
 
+-- | Handle the common case of fetching Profile information a JSON endpoint
+--
+-- Throws @'InvalidProfileResponse'@ if JSON parsing fails
+--
+handleProfileResponse :: FromJSON a
```
jferris: Is there a better word we can use here than "handle?" Map? Decode?

pbrisbin: I wouldn't say it does either of those things. It's already decoded when we get the `a` and `map` is very overloaded. Maybe "convert"? I chose "handle" for "error handling".

jferris: Maybe `tryAuth' or `authOrFail`?

pbrisbin: I'm not used to worrying about naming so much in Haskell. Can I just call it `f`?

How about `fromProfileURL`?

jferris: Haha, `f`.

Yeah, I think `fromProfileURL` is pretty good.

pbrisbin:
Ready to merge with that rename?

jferris:
Looks good!

PR #33, opened 2015-06-27
Added EVE-Online SSO and made it possible for plugins to use images …

Drezil:
…instead of text-links to login

- bumped version to 0.1.3
- added EveOnline SSO
- added authOAuth2Image, while keeping authOAuth2 backward-compatible
- Plugins can now use authOAuth2Image to ask for and provide images (Either local or remote)

### DIFF
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -11,6 +11,7 @@
 --
 module Yesod.Auth.OAuth2
     ( authOAuth2
+    , authOAuth2Image
     , oauth2Url
     , fromProfileURL
     , YesodOAuth2Exception(..)
@@ -57,9 +58,22 @@ authOAuth2 :: YesodAuth m
            --   second authorized request to @api/me.json@.
            --
            --   See @'fromProfileURL'@ for an example.
-           --
            -> AuthPlugin m
-authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
+authOAuth2 name oauth getCreds = authOAuth2Image name oauth getCreds Nothing
+
+authOAuth2Image :: YesodAuth m
+           => Text   -- ^ Service name
+           -> OAuth2 -- ^ Service details
+           -> (Manager -> AccessToken -> IO (Creds m))
+           -- ^ This function defines how to take an @'AccessToken'@ and
+           --   retrieve additional information about the user, to be
+           --   set in the session as @'Creds'@. Usually this means a
+           --   second authorized request to @api/me.json@.
+           --
+           --   See @'fromProfileURL'@ for an example.
+           -> Maybe Text -- ^ URL to image shown instead of "Login with xxx"-Text
+           -> AuthPlugin m
+authOAuth2Image name oauth getCreds im = AuthPlugin name dispatch login
 
   where
     url = PluginR name ["callback"]
@@ -105,7 +119,11 @@ authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
     tokenSessionKey = "_yesod_oauth2_" <> name
 
     login tm = [whamlet|
-        <a href=@{tm $ oauth2Url name}>Login via #{name}
+        <a href=@{tm $ oauth2Url name}>
+          $maybe image <- im
+            <img src=#{image}>
+          $nothing
+            Login via #{name}
         |]
 
 -- | Handle the common case of fetching Profile information a JSON endpoint
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -11,7 +11,7 @@
 --
 module Yesod.Auth.OAuth2
     ( authOAuth2
-    , authOAuth2Image
+    , authOAuth2Widget
     , oauth2Url
     , fromProfileURL
     , YesodOAuth2Exception(..)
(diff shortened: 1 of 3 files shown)


Drezil:
will fix build errores and PR again.

yesod-auth-oauth2.cabal
```diff
@@ -1,5 +1,5 @@
 name:            yesod-auth-oauth2
-version:         0.1.2
+version:         0.1.3
```
pbrisbin: Can you revert this? Version/release is a separate concern from any one PR/feature.

yesod-auth-oauth2.cabal
```diff
@@ -36,6 +36,8 @@ library
                    , transformers            >= 0.2.2     && < 0.5
                    , hoauth2                 >= 0.4.7     && < 0.5
                    , lifted-base             >= 0.2       && < 0.4
+                   , time                    
+                   --dependency of yesod-core. Let cabal choose version.
```
pbrisbin: Can you explain the reasoning for this a little bit more?

pbrisbin:
Thanks for this. The new plugin itself looks great and I'm happy to merge.

For the image handling, I wonder if it would be more extensible and simpler to add some attributes to all the login markup allowing users to target the links with CSS `background-urls` if they want images instead of text. Would that work for your use case?

Yesod/Auth/OAuth2/EveOnline.hs
```diff
@@ -0,0 +1,115 @@
...
+             -> ImageType
+             -> AuthPlugin m
+oauth2EveImage clientId clientSecret im = oauth2EveScoped clientId clientSecret ["publicData"] (Just . toURI $ im)
+             where
+               toURI :: ImageType -> Text
+               toURI BigWhite = "https://images.contentful.com/idjq7aai9ylm/4PTzeiAshqiM8osU2giO0Y/5cc4cb60bac52422da2e45db87b6819c/EVE_SSO_Login_Buttons_Large_White.png?w=270&h=45"
+               toURI BigBlack = "https://images.contentful.com/idjq7aai9ylm/4fSjj56uD6CYwYyus4KmES/4f6385c91e6de56274d99496e6adebab/EVE_SSO_Login_Buttons_Large_Black.png?w=270&h=45"
+               toURI SmallWhite = "https://images.contentful.com/idjq7aai9ylm/18BxKSXCymyqY4QKo8KwKe/c2bdded6118472dd587c8107f24104d7/EVE_SSO_Login_Buttons_Small_White.png?w=195&h=30"
+               toURI SmallBlack = "https://images.contentful.com/idjq7aai9ylm/12vrPsIMBQi28QwCGOAqGk/33234da7672c6b0cdca394fc8e0b1c2b/EVE_SSO_Login_Buttons_Small_Black.png?w=195&h=30"
+               toURI (Custom a) = a
```
pbrisbin: Can you dedent this to match the project style? The `where` should be two spaces in and the sub-clauses  four.

Yesod/Auth/OAuth2/EveOnline.hs
```diff
@@ -0,0 +1,115 @@
...
+import Data.Text as T (Text,unwords)
+import Data.ByteString as B (ByteString)
+import Data.Text.Encoding (encodeUtf8, decodeUtf8)
+import Network.HTTP.Conduit (Manager)
+import Yesod.Auth
+import Yesod.Auth.OAuth2
+
+import qualified Data.Text as T
+
+data ImageType = BigWhite
```
pbrisbin: To match project style, can you format this as:

```hs
data ImageType
    = BigWhite
    | ...
    | ...
```

?

Yesod/Auth/OAuth2/EveOnline.hs
```diff
@@ -0,0 +1,115 @@
...
+        <*> (T.words <$> o .: "Scopes")
+        <*> o .: "TokenType"
+        <*> o .: "CharacterOwnerHash"
+
+    parseJSON _ = mzero
+
+oauth2Eve :: YesodAuth m
+             => Text -- ^ Client ID
+             -> Text -- ^ Client Secret
+             -> AuthPlugin m
```
pbrisbin: Can you align the `=>`/`->` with the `::`?

Yesod/Auth/OAuth2/EveOnline.hs
```diff
@@ -0,0 +1,115 @@
...
+oauth2Eve clientId clientSecret = oauth2EveScoped clientId clientSecret ["publicData"] Nothing
+
+oauth2EveImage :: YesodAuth m
+             => Text -- ^ Client ID
+             -> Text -- ^ Client Secret
+             -> ImageType
+             -> AuthPlugin m
```
pbrisbin: Here as well.

Drezil:
That could work - although many users would have to write the same code for the same images in e.g. `/static/auth/(github|twitter|eveonline|...)-login.png`.
The solution i included was to give a URL to the login-image. It could be changed to a widget (with complete custom whamlet/lucius-code) easily - and be backward-compatible with the text-only-login-link..

pbrisbin:
> although many users would have to write the same code for the same images in e.g. /static/auth/(github|twitter|eveonline|...)-login.png.

That's true, but is an existing situation, not a downside introduced by the approach.  Users adding images for these plugins today already have a tough time (the only option is overriding `loginHandler` entirely), this would make easier for them.

> It could be changed to a widget (with complete custom whamlet/lucius-code) easily - and be backward-compatible with the text-only-login-link..

This sounds better, would you be willing to code it up that way in this PR?

Yesod/Auth/OAuth2.hs
```diff
@@ -57,9 +58,22 @@ authOAuth2 :: YesodAuth m
...
+           -> OAuth2 -- ^ Service details
+           -> (Manager -> AccessToken -> IO (Creds m))
+           -- ^ This function defines how to take an @'AccessToken'@ and
+           --   retrieve additional information about the user, to be
+           --   set in the session as @'Creds'@. Usually this means a
+           --   second authorized request to @api/me.json@.
            --
+           --   See @'fromProfileURL'@ for an example.
+           -> Maybe Text -- ^ URL to image shown instead of "Login with xxx"-Text
            -> AuthPlugin m
```
pbrisbin: Sorry, one more `::`-alignment fix.

(Thanks for being accommodating here)

Drezil:
Now you can put in any widget with OAuth2Widget or just use the default text-link with OAuth2.

For the other plugins there should be something similar with defaults either in the web or at `static/auth/pluginname.png` for login, so you can put your image there..

Yesod/Auth/OAuth2.hs
```diff
@@ -57,9 +58,22 @@ authOAuth2 :: YesodAuth m
            --   second authorized request to @api/me.json@.
            --
            --   See @'fromProfileURL'@ for an example.
-           --
            -> AuthPlugin m
-authOAuth2 name oauth getCreds = AuthPlugin name dispatch login
+authOAuth2 name oauth getCreds = authOAuth2Widget name oauth getCreds Nothing
```
pbrisbin: Instead of using a `Maybe`, how about we just pass `[whamlet|Login via #{name}|]` here?

Drezil:
I just wanted to say that we dont know #{name} there .. but we do ..

i will change that quickly. That maybe was bothering me anyway :)

pbrisbin:
We do, but I'm not sure if there's value in passing `\name -> [whamlet|Login via #{name}|]` anyway... would a call ever not have name? Would `authOAuth2Widget` ever pass a different name than the caller has?

Drezil:
I dont know. I would just give the freedom to the user of the library. If they want it differently, they can have it.

pbrisbin:
LGTM -- will merge and release soon. Thanks!

PR #34, opened 2015-06-29
Prep for release after Eve Online merge

pbrisbin:
Fixes to EveOnline plugin

- Fix invalid data declaration
- Fix -Wall compilation warnings
  - Unused imports
  - Unused functions
  - Unused arguments
  - Shadowed function name
- Incorporate plain link into WidgetType, remove Maybe
- Use only one function, always accept WidgetType

Add/fix documentation, change argument order

- Document the difference between widget and non-widget functions
- Pass the widget as the first argument to take advantage of partial
  application

/cc @Drezil

### DIFF
--- a/Yesod/Auth/OAuth2/EveOnline.hs
+++ b/Yesod/Auth/OAuth2/EveOnline.hs
@@ -7,7 +7,7 @@
 --
 -- * Authenticates against eveonline
 -- * Uses EVEs unique account-user-char-hash as credentials identifier
--- * Returns charName, tokenType and expires as extras
+-- * Returns charName, tokenType, accessToken and expires as extras
 --
 module Yesod.Auth.OAuth2.EveOnline
     ( oauth2Eve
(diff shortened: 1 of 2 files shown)


Yesod/Auth/OAuth2/EveOnline.hs
```diff
@@ -35,18 +33,19 @@ import Yesod.Core.Widget
...
+    = Plain -- ^ Simple "Login via eveonline" text
+    | BigWhite
     | SmallWhite
     | BigBlack
     | SmallBlack
     | Custom (WidgetT m IO ())
 
 data EveUser = EveUser
-    { eveUserId :: Int
+    { _eveUserId :: Int
```
jferris: Was this shadowing something? If the field is unused, can we drop it entirely?

pbrisbin:
The were being parsed out of the JSON but otherwise unused (hence the warning). I've removed them entirely now.

pbrisbin:
Ping @jferris any reason not to merge this?

jferris:
Nope - looks good to merge.

PR #37, opened 2015-08-10
New provider for Nylas

bts:
Hi there,

Thanks for this project. I've implemented a new provider for [Nylas](https://nylas.com/).

I've decided to use the unique ID for the [namespace](https://nylas.com/docs/platform/api_namespaces) that Nylas generates for each account as the `credsIdent`. This is because Nylas performs a "canonicalization" transform on email address it returns, potentially resulting in a mismatch with what a client might already have on-file for a user.

Additionally, I had to add an explicit dependency on `http-client` because Nylas asks for [the access token to be supplied as the Basic Auth username](https://nylas.com/docs/platform/authentication) when fetching credentials (so we couldn't just use `authGetJSON`).

Let me know if you have any questions, or if I can improve anything (I'm new to Haskell).

Thanks!
Brian

### DIFF
--- a/Yesod/Auth/OAuth2/Nylas.hs
+++ b/Yesod/Auth/OAuth2/Nylas.hs
@@ -53,15 +53,14 @@ oauth2Nylas :: YesodAuth m
             => Text -- ^ Client ID
             -> Text -- ^ Client Secret
             -> AuthPlugin m
-oauth2Nylas clientId clientSecret =
-    oauth2NylasScoped clientId clientSecret ["email"]
+oauth2Nylas = oauth2NylasScoped ["email"]
 
 oauth2NylasScoped :: YesodAuth m
-                  => Text -- ^ Client ID
-                  -> Text -- ^ Client Secret
-                  -> [Text] -- ^ Scopes
+                  => [Text] -- ^ Scopes
+                  -> Text   -- ^ Client ID
+                  -> Text   -- ^ Client Secret
                   -> AuthPlugin m
-oauth2NylasScoped clientId clientSecret scopes =
+oauth2NylasScoped scopes clientId clientSecret =
     authOAuth2 "nylas" oauth fetchCreds
   where
     authorizeUrl = encodeUtf8
--- a/Yesod/Auth/OAuth2/Nylas.hs
+++ b/Yesod/Auth/OAuth2/Nylas.hs
@@ -17,11 +17,13 @@ import Data.Monoid ((<>))
 import Data.Text (Text)
 import Data.Text.Encoding (encodeUtf8, decodeUtf8)
 import Data.Vector ((!?))
-import Network.HTTP.Client (applyBasicAuth, parseUrl, httpLbs, responseStatus, responseBody)
+import Network.HTTP.Client (applyBasicAuth, parseUrl, httpLbs, responseStatus
+                           , responseBody)
 import Network.HTTP.Conduit (Manager)
-
-import Yesod.Auth
-import Yesod.Auth.OAuth2
+import Yesod.Auth (Creds(..), YesodAuth, AuthPlugin)
+import Yesod.Auth.OAuth2 (OAuth2(..), AccessToken(..)
+                         , YesodOAuth2Exception(InvalidProfileResponse)
+                         , authOAuth2)
 
 import qualified Data.Text as T
 import qualified Network.HTTP.Types as HT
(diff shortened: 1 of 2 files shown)


Yesod/Auth/OAuth2/Nylas.hs
```diff
@@ -0,0 +1,93 @@
...
+    { nylasNamespaceId :: Text
+    , nylasNamespaceAccountId :: Text
+    , nylasNamespaceEmailAddress :: Text
+    , nylasNamespaceName :: Text
+    , nylasNamespaceProvider :: Text
+    , nylasNamespaceOrganizationUnit :: Text
+    }
+
+instance FromJSON NylasNamespace where
+    parseJSON (Array singleton) = case singleton !? 0 of
```
pbrisbin: You could incorporate the `case` into the pattern match directly I think:

```hs
parseJSON (Array (Object o:_)) = NylasNamespace
  -- ...
parseJSON _ = mzero
```

pbrisbin:
One minor comment, otherwise this looks great to me. Thanks!

bts:
Is there some way to do this kind of pattern-match on a `Vector Value` (instead of a `[Value]`)?

```
Yesod/Auth/OAuth2/Nylas.hs:41:23-32: Couldn't match type ‘[Value]’ with ‘Data.Vector.Vector Value’ …
    Expected type: aeson-0.8.0.2:Data.Aeson.Types.Internal.Array
      Actual type: [Value]
    In the pattern: Object o : _
    In the pattern: Array (Object o : _)
    In an equation for ‘parseJSON’:
        parseJSON (Array (Object o : _))
          = NylasNamespace <$> o .: "id" <*> o .: "account_id"
            <*> o .: "email_address"
            <*> o .: "name"
            <*> o .: "provider"
            <*> o .: "organization_unit"
```

pbrisbin:
Oh, I didn't realize it wasn't a normal list. It looks like with `Vector` the way you had it is the only way to do it. Even it were constructed such that you could pattern match the head out, the constructors don't seem to be exported anyway.

pbrisbin:
Ping @jferris any concerns? If not, I can probably merge and release some time this week.

Yesod/Auth/OAuth2/Nylas.hs
```diff
@@ -0,0 +1,93 @@
...
+            <*> o .: "account_id"
+            <*> o .: "email_address"
+            <*> o .: "name"
+            <*> o .: "provider"
+            <*> o .: "organization_unit"
+        _ -> mzero
+    parseJSON _ = mzero
+
+oauth2Nylas :: YesodAuth m
+            => Text -- ^ Client ID
```
jferris: Thoughts on making a `type` alias for `ClientID`/`ClientSecret` to replace these comments?

Yesod/Auth/OAuth2/Nylas.hs
```diff
@@ -0,0 +1,93 @@
...
+                  => Text -- ^ Client ID
+                  -> Text -- ^ Client Secret
+                  -> [Text] -- ^ Scopes
+                  -> AuthPlugin m
+oauth2NylasScoped clientId clientSecret scopes = authOAuth2 "nylas" oauth fetchCreds
+  where
+    oauth = OAuth2
+        { oauthClientId = encodeUtf8 clientId
+        , oauthClientSecret = encodeUtf8 clientSecret
+        , oauthOAuthorizeEndpoint = encodeUtf8 $ "https://api.nylas.com/oauth/authorize?scope=" <> T.intercalate "," scopes
```
jferris: Could we wrap this line at 80 characters?

Worth extracting a `where` binding for the URL value?

jferris:
@pbrisbin I had a couple formatting/clarity suggestions, but nothing major.

Yesod/Auth/OAuth2/Nylas.hs
pbrisbin: I agree. Do you think that'd be better as an overall change separately though? The `type`s could be provided by the main module and all the plugins have similar functions that could use them.

Yesod/Auth/OAuth2/Nylas.hs
pbrisbin: I'm fine with a `where` clause or just breaking at the `$` (I forget if we have a guideline about before/after operators for these breaks).

Yesod/Auth/OAuth2/Nylas.hs
jferris: That makes sense to me. We might also be able to make `newtype`s for these to avoid mistakes.

bts:
Thanks for the feedback! Let me know if there's anything else I can tweak.

pbrisbin:
This LGTM -- I'll try to get it released today.

pbrisbin:
Rebased and merged.

bts:
Thanks!

PR #38, opened 2015-08-12
Bump upper bounds on aeson

lukerandall:
This is holding back the aeson upgrade in stackage (https://github.com/fpco/stackage/issues/572)


pbrisbin:
LGTM -- will get try to get this out today

PR #39, opened 2015-08-20
Switch to newer Nylas API endpoint

bts:
Hi there,

Nylas just updated their API today, and the existing endpoint (`/n`, for "namespace") we were using will unfortunately be decommissioned sometime after October 1, 2015. This patch switches to their new `/account` endpoint.

There are no changes to the types of the user-facing `oauth2Nylas` or `oauth2NylasScoped` functions, but we now no longer return one of the entries in `credsExtra`: what Nylas _used to_ call the `account_id`, but wasn't actually used for anything. Nylas has stopped serving this older notion of `account_id`, and now returns what was called the `namespace_id` as `account_id`. We're still returning the user this same value as our `credsIdent`.

Let me know if you have any questions.

Thanks!
Brian

### DIFF
--- a/Yesod/Auth/OAuth2/Nylas.hs
+++ b/Yesod/Auth/OAuth2/Nylas.hs
@@ -16,7 +16,6 @@ import Data.Aeson (FromJSON, Value(..), parseJSON, decode, (.:))
 import Data.Monoid ((<>))
 import Data.Text (Text)
 import Data.Text.Encoding (encodeUtf8, decodeUtf8)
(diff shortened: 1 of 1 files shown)


jferris:
This looks good to me.

@pbrisbin any objections to merging this in?

pbrisbin:
Nope. Go for it.

jferris:
Thanks! I merged this into master.

bts:
Thank you!

PR #41, opened 2015-10-03
Bump upper bound on vector to <11

### LINKED ISSUE #40: Build against vector-0.11

pbrisbin:
Fixes #40

/cc @jferris

I'm not sure what testing to do, I've confirmed it compiles.


jferris:
I am comfortable with assuming this is good given that it compiles.

PR #45, opened 2015-11-28
Append the state query parameter correctly

### LINKED ISSUE #44: Adding state parameter causes malformed URL
I'm building a simple intgration for LinkedIn. Everything is working except that there it generates a slightly malformed URL. The authorize endpoint ends up looking like this:

    https://www.linkedin.com/uas/oauth2/authorization&state=rfvjuhwswetbcdwzfeoghnxhovhjjz?client_id=7730si35gvu8rr&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fauth%2Fpage%2Flinkedin%2Fcallback

The ampersand after the word `authorization` needs to be a question mark, and the question mark after the state parameter needs to be an ampersand.

I've tracked the source of the error down to this piece of the code: https://github.com/thoughtbot/yesod-auth-oauth2/blob/4afaba6645a0df69c4cf0760ff7ad56c69057bc2/Yesod/Auth/OAuth2.hs#L89-L90

My code that uses the library looks like this:

    oauth2LinkedIn clientId clientSecret = authOAuth2 "linkedin"
        OAuth2
            { oauthClientId = encodeUtf8 clientId
            , oauthClientSecret = encodeUtf8 clientSecret
            , oauthOAuthorizeEndpoint = "https://www.linkedin.com/uas/oauth2/authorization"
            , oauthAccessTokenEndpoint = "https://www.linkedin.com/uas/oauth2/accessToken"
            , oauthCallback = Nothing
            }
        $ fromProfileURL
            "linkedin"
            "https://api.linkedin.com/v1/people/~?format=json"
            credsFromLinkedInUser

pbrisbin:
If the endpoint URL doesn't yet have any query parameters, this code would
construct an invalid URL by appending `&state=` rather than `?state=`. We now
correctly append `?state=` or `&state=` depending on the URL already containing
a `?` character.

Fixes #44

This is a "good enough" solution for the moment, done for ease of
implementation and speed of delivery. An eventual better solution would be to
use a URL-parsing and constructing library to parse the existing endpoint,
append parameter tuples, and re-construct a new URL.

Restructuring the data types so we start handling real URL and QueryParam
values instead of ByteStrings would also be an option, though we're a little
cornered by the Yesod.Auth.OAuth interface.

/cc @thoughtbot/haskell @jferris

### DIFF
--- a/Yesod/Auth/OAuth2.hs
+++ b/Yesod/Auth/OAuth2.hs
@@ -38,6 +38,7 @@ import Yesod.Core
 import Yesod.Form
 
 import qualified Data.ByteString.Lazy as BL
+import qualified Data.ByteString.Char8 as C8
 
 -- | Provider name and Aeson parse error
 data YesodOAuth2Exception = InvalidProfileResponse Text BL.ByteString
@@ -87,7 +88,7 @@ authOAuth2Widget widget name oauth getCreds = AuthPlugin name dispatch login
         return oauth
             { oauthCallback = Just $ encodeUtf8 $ render $ tm url
             , oauthOAuthorizeEndpoint = oauthOAuthorizeEndpoint oauth
-                <> "&state=" <> encodeUtf8 csrfToken
+                `appendQuery` "state=" <> encodeUtf8 csrfToken
             }
 
     dispatch "GET" ["forward"] = do
@@ -141,3 +142,9 @@ fromProfileURL name url toCreds manager token = do
 
 bsToText :: ByteString -> Text
 bsToText = decodeUtf8With lenientDecode
+
+appendQuery :: ByteString -> ByteString -> ByteString
+appendQuery url query =
+    if '?' `C8.elem` url
+        then url <> "&" <> query
+        else url <> "?" <> query


jferris:
The quick solution looks good to me. Is this worth having a test somewhere for?

> An eventual better solution would be to use a URL-parsing and constructing library to parse the existing endpoint, append parameter tuples, and re-construct a new URL.

Agreed, but then we'd have to handle the case where the URL fails to parse, right?

> Restructuring the data types so we start handling real URL and QueryParam values instead of ByteStrings would also be an option, though we're a little cornered by the Yesod.Auth.OAuth interface.

This sounds like it would be a solid choice, since we'd parse the URL early and know for sure we had a valid one by the time we started appending parameters.

pbrisbin:
I can spend some time seeing if I can get a test around this, but with no testing harness in place yet, and the relatively small bit of logic I'm aiming to test pretty deep behind the public interface, I'm not sure it's viable.

jferris:
Okay. Sounds like "get a test harness in place" is another good TODO.

pbrisbin:
Adding a test harness was super trivial, but actually testing any aspect of the `AuthPlugin` values we're working with is shaping up to be impossible.

I'm going to merge this as-is and ship to get this bugfix out, but I'll open a new PR with my WIP testing stuff for us to keep poking at.

PR #46, opened 2015-12-05
Add potential, currently non-functioning tests

pbrisbin:
Approach so far:

The SpecHelper sets up an example App type which should (theoretically) be
enough to get some fake requests and responses going, as in any
yesod-test-based suite.

The spec then tries to build an example plugin and make assertions on how it
dispatches. This is currently falling down on subsite-related type errors (see
commented attempts).

Another potential direction is to define the YesodAuth instance for App to
specify authPlugins built using the library. With that, we might be able to
create specs using yesod-test that exercise aspects of the plugins in an
indirect way, but enough to make useful assertions.

/cc @jferris

### DIFF
--- a/test/SpecHelper.hs
+++ b/test/SpecHelper.hs
@@ -1,0 +1,43 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+module SpecHelper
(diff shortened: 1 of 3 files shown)


silky:
(it'd be good to merge this; so that one can write some tests, but also to get this under stack.)

pbrisbin:
I can definitely pull out stack and a no-op testing target out. Thanks for the bump.

pbrisbin:
I don't know for sure, but I'd suspect getting something like what I was poking at working will look different enough from this PR that its only value is historical record of the attempt -- which it retains even after being closed. So I'm closing it.

If anyone wants to pick up the testing torch (with this or a new branch), please feel free.

PR #50, opened 2016-01-02
Properly capitalise GitHub link

silky:
currently the generated link says "Login with github"; it'd be nicer if it said "Login with GitHub". this pr makes it so.

### DIFF
--- a/Yesod/Auth/OAuth2/Github.hs
+++ b/Yesod/Auth/OAuth2/Github.hs
@@ -67,7 +67,7 @@ oauth2GithubScoped :: YesodAuth m
              -> Text -- ^ Client Secret
              -> [Text] -- ^ List of scopes to request
              -> AuthPlugin m
-oauth2GithubScoped clientId clientSecret scopes = authOAuth2 "github" oauth fetchGithubProfile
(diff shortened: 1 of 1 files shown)


pbrisbin:
Unfortunately, I think I'm against this. If there are presentation-level concerns such as this, I think it should be left to the application. Here's how we do it in [carnvial](https://github.com/thoughtbot/carnival/blob/master/templates/login.hamlet#L5-L6).

You'll also notice that that code is an example where end-users are relying on the plugin's name (`"github"`) being stable. It's unfortunate we don't have type-level safety for that, but it does indicate that a change like this has wider impact (i.e. risk) than we might think.

WDYT @jferris ?

silky:
alright, i understand

it's a shame, because the i currently grab out the auth list by calling `apLogin` on all the plugins in `authPlugins master`, and it comes back nicely for the GoogleAuth2 one, that is in yesod; and it means i don't need to hardcode the list

(i have exactly the same sitation with the `dummy` login).

but i can accept not changing it; given that it is a breaking change. it's not a huge deal.

pbrisbin:
Yeah, I wish we had thought to use `"GitHub"` originally. Unfortunately, I think the compatibility concerns are too much friction for me right now. Thanks for understanding.

PR #52, opened 2016-01-03
Use primary email as email, if one is set.

### LINKED ISSUE #51: select 'primary' email on github email list
currently the code takes the first email in the list; but it should take the one with `primary: true`.

see - https://developer.github.com/v3/users/emails/

silky:
If we couldn't find a primary email, use the first email returned from the
list. This fixes #51.

This is might be a breaking change, though, in that if this selects a different email address for anyone, and they are already registered, this field will be different.

### DIFF
--- a/Yesod/Auth/OAuth2/Github.hs
+++ b/Yesod/Auth/OAuth2/Github.hs
@@ -21,6 +21,8 @@ import Control.Applicative ((<$>), (<*>))
 import Control.Exception.Lifted
 import Control.Monad (mzero)
 import Data.Aeson
+import Data.Maybe (fromMaybe)
+import Data.List (find)
 import Data.Monoid ((<>))
 import Data.Text (Text)
 import Data.Text.Encoding (encodeUtf8, decodeUtf8)
@@ -49,12 +51,14 @@ instance FromJSON GithubUser where
     parseJSON _ = mzero
 
 data GithubUserEmail = GithubUserEmail
-    { githubUserEmail :: Text
+    { githubUserEmailAddress :: Text
+    , githubUserEmailPrimary :: Bool
     }
 
 instance FromJSON GithubUserEmail where
     parseJSON (Object o) = GithubUserEmail
         <$> o .: "email"
+        <*> o .: "primary"
 
     parseJSON _ = mzero
 
@@ -91,11 +95,11 @@ fetchGithubProfile manager token = do
         (_, Left err) -> throwIO $ InvalidProfileResponse "github" err
 
 toCreds :: GithubUser -> [GithubUserEmail] -> AccessToken -> Creds m
-toCreds user userMail token = Creds
+toCreds user userMails token = Creds
     { credsPlugin = "github"
     , credsIdent = T.pack $ show $ githubUserId user
     , credsExtra =
-        [ ("email", githubUserEmail $ head userMail)
+        [ ("email", githubUserEmailAddress email)
         , ("login", githubUserLogin user)
         , ("avatar_url", githubUserAvatarUrl user)
         , ("location", githubUserLocation user)
@@ -104,5 +108,6 @@ toCreds user userMail token = Creds
     }
 
   where
+    email = fromMaybe (head userMails) $ find githubUserEmailPrimary userMails
     maybeName Nothing = []
     maybeName (Just name) = [("name", name)]


Yesod/Auth/OAuth2/Github.hs
```diff
@@ -101,5 +104,8 @@ toCreds user userMail token = Creds
     }
 
   where
+    email = githubUserEmailEmail $ case find githubUserEmailPrimary userMails of
+                   Nothing -> head userMails
+                   Just e  -> e
```
pbrisbin: I think this might be clearer with `fromMaybe`

pbrisbin:
One note, LGTM besides. I'll wait for that, #55, rebase and merge.

pbrisbin:
> This is might be a breaking change, though, in that if this selects a different email address for anyone, and they are already registered, this field will be different.

I just noticed this. Definitely worth some thought. WDYT @jferris ?

pbrisbin:
Stack / Circle CI has merged. If you have a moment to rebase this, that'd be great. I didn't want to cut a new, rebased PR since we do have a meaningful ongoing discussion here IMO.

Yesod/Auth/OAuth2/Github.hs
```diff
@@ -47,12 +49,14 @@ instance FromJSON GithubUser where
     parseJSON _ = mzero
 
 data GithubUserEmail = GithubUserEmail
-    { githubUserEmail :: Text
+    { githubUserEmailEmail :: Text
```
jferris: Thoughts on `EmailAddress` or `EmailText` vs `EmailEmail` here?

jferris:
The unique key is based on the `credsIdent` field, right? That means that it will be based on GitHub user ID and not email, so this shouldn't *break* anything. It will return the primary address instead of an arbitrary one, which I'd say is a bug fix.

silky:
@jferris yes one would typically use the `credsIdent` field (along with the name of the plugin) for the unique user lookup; so this shouldn't break anything there.

i guess i just wanted to highlight that this may return a different email address associated with the auth, and that consumers of the plugin may want to be aware of that.

Yesod/Auth/OAuth2/Github.hs
silky: regarding this, the current name is crazy i agree, but i did that just so it would be consistent.

i thought that a better win would be to build up a better record structure like say - https://github.com/silky/yesod/blob/master/yesod-auth/Yesod/Auth/GoogleEmail2.hs#L470 - and then just solve this in that refactoring (i'm happy to make this change, but probs in a different PR).

pbrisbin:
Seems reasonable enough to me.

silky:
@pbrisbin i've rebased this; let me know if there's something more to do.

pbrisbin:
I'll let you and @jferris come to an agreement on `githubEmailEmail` (I don't have a strong opinion). Once that's resolved (or deferred), I'd say this is mergable.

silky:
@jferris bump

Yesod/Auth/OAuth2/Github.hs
jferris: Could you show the better record structure you're thinking of?

silky: fair enough; thinking about it, it'd pretty much what is there but the GithubUser would have a field `emails :: [GithubUserEmail]`, and the `GithubUserEmail` would be exactly what's there and probably instead of `..EmailEmail` it'd be better as `...EmailAddress`.

so yeah, i agree `EmailAddress` is better than what i have. happy to change it

jferris: > `GithubUser` would have a field `emails :: [GithubUserEmail]`

I think this would be cool, but it's a little tricky because the basic profile info that's currently in `GithubUser` comes from a different web request as the emails. I'd be fine with making this change, but I think it would be best to do in another pull request.

I think the `GithubUserEmailEmail` to `GithubUserEmailAddress` changes would be good to make. If you'd like to make those, I'd be happy to merge this in.

silky:
updated with `EmailEmail -> EmailAddress` change. sorry it took me so long!

jferris:
No worries, thanks for the change! I merged this into master.

PR #55, opened 2016-01-03
Stack, Circle CI, and a no-op Hspec suite

pbrisbin:
Getting under stack make sense to me, but the changes here don't preclude other
development approaches. Nothing's actually tested, but at least it's now
obvious how we'll want to test when we can figure out an approach. The Travis
build is currently broken, so rather than fight with it I decided to just use a
simpler, stack-based Circle CI setup that I've used in other projects.

See individual commit messages for additional details.

/cc @jferris @silky


circle.yml
```diff
@@ -0,0 +1,14 @@
+dependencies:
+  cache_directories:
+    - "~/.stack"
+  pre:
```
jferris: Will this run every time, or will it get cached between builds? If it runs every time, do we want to download this into a directory that gets cached?

jferris:
Nice. I had a question about caching, but this approach looks good to me.

circle.yml
pbrisbin: The steps run every time. The ones here in `pre` take almost no time so I didn't think it was super important to put their artifacts into `~/.stack` (or some other cached directory). The `stack setup` and `stack build` steps are the time-consuming ones and their artifacts are what are cached via `~/.stack`.

You can see example timings and outcomes [here](https://circleci.com/gh/pbrisbin/tee-io/79). For example, the `stack build` step only compiles the actual application code, all dependencies were pre-installed because they live in `~/.stack` (presumably).

jferris: Cool. I assumed that downloading Stack would take more than a couple seconds, but caching isn't worth it given your example.

PR #58, opened 2016-02-07
Remove Twitter module

### LINKED ISSUE #25: Add OAuth2.Twitter

pbrisbin:
- Twitter [does not support OAuth 2][1]

  The motivating issue was created without realizing this (it is a bit
  surprising), and the module that was committed doesn't appear to work.

- For OAuth 1, [support][2] already exists

Fixes #25

[1]: https://twittercommunity.com/t/oauth-2-0-support/253
[2]: http://hackage.haskell.org/package/yesod-auth-oauth-1.4.0.2/docs/Yesod-Auth-OAuth.html

/cc @jferris @gabebw


gabebw:
:+1:

jferris:
I guess this makes sense. It looks like Twitter added [support for Oauth2](https://dev.twitter.com/oauth/reference/post/oauth2/token), but not for user authentication. For Yesod-based auth, I don't think that makes sense, so using Oauth1 is the only way to go.

PR #59, opened 2016-02-07
Add example application

pbrisbin:
/cc @jferris

This could also be used as a pre-release smoke test since adding more
traditional tests has proved difficult.

### DIFF
--- a/yesod-auth-oauth2.cabal
+++ b/yesod-auth-oauth2.cabal
@@ -16,6 +16,10 @@ flag network-uri
    description: Get Network.URI from the network-uri package
    default: True
 
+flag example
+  description: Build the example application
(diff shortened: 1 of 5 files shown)


yesod-auth-oauth2.cabal
```diff
@@ -16,6 +16,10 @@ flag network-uri
    description: Get Network.URI from the network-uri package
    default: True
 
+flag example
```
jferris: Neat.

jferris:
This looks good to me.

Worth changing the CI config to make sure the example app builds?

pbrisbin:
Yeah, I think so. I'll do that then merge on green.

PR #60, opened 2016-02-07
Include public_email key in credsExtra for Github

### LINKED ISSUE #54: consider checking public github email address then falling back to /user/email
github returns the "public" address via `/users/:username`, and if this is `null` we could fall back to the current method of returning the first primary or just first email.

some users (in particular, me) have a different public email address set then the ones that come back from `/user/emails`.

see - https://developer.github.com/v3/users/

pbrisbin:
Fixes #54


silky:
looks fine to me!

jferris:
Looks good to me, too.

PR #63, opened 2016-06-09
Relax bound on transformers for GHC 8

paul-rouse:
I've tested this change against stackage nightly-2016-06-08 using a small example (not the one in the repo, which seems to fall foul of https://ghc.haskell.org/trac/ghc/ticket/12130, but similar).

When released, I am quite happy to PR the re-enabling of this package in stackage.


jferris:
Thanks! I merged this into master.

Note: I ended up splitting out the version bump into its own commit, and went with 0.1.9.

PR #64, opened 2016-06-20
Properly handle empty "location" fields.

adarqui:
Hi everyone,

This PR simply makes "location" optional. I recently configured a secondary github account to test my application with two different users. The new github account lacked several fields, including "location". Without this change, I would receive the following error:

```
20/Jun/2016:01:59:20 -0400 [Warn] Error Resp: InternalError "InvalidProfileResponse \"github\" \"hoauth2.HttpClient.parseResponseJSON/Could not decode JSON:
...
```

Upon further inspection, "location" was null. After making these changes and testing locally, I was able to successfully register via Github.

Thanks alot!

-- Andrew

### DIFF
--- a/Yesod/Auth/OAuth2/Github.hs
+++ b/Yesod/Auth/OAuth2/Github.hs
@@ -37,7 +37,7 @@ data GithubUser = GithubUser
     , githubUserName :: Maybe Text
     , githubUserLogin :: Text
     , githubUserAvatarUrl :: Text
-    , githubUserLocation :: Text
+    , githubUserLocation :: Maybe Text
     , githubUserPublicEmail :: Maybe Text
     }
 
@@ -47,8 +47,8 @@ instance FromJSON GithubUser where
         <*> o .:? "name"
         <*> o .: "login"
         <*> o .: "avatar_url"
-        <*> o .: "location"
-        <*> o .: "email"
+        <*> o .:? "location"
+        <*> o .:? "email"
 
     parseJSON _ = mzero
 
@@ -104,11 +104,11 @@ toCreds user userMails token = Creds
         [ ("email", githubUserEmailAddress email)
         , ("login", githubUserLogin user)
         , ("avatar_url", githubUserAvatarUrl user)
-        , ("location", githubUserLocation user)
         , ("access_token", decodeUtf8 $ accessToken token)
         ]
         ++ maybeName (githubUserName user)
         ++ maybePublicEmail (githubUserPublicEmail user)
+        ++ maybeLocation (githubUserLocation user)
     }
 
   where
@@ -119,3 +119,6 @@ toCreds user userMails token = Creds
 
     maybePublicEmail Nothing = []
     maybePublicEmail (Just e) = [("public_email", e)]
+
+    maybeLocation Nothing = []
+    maybeLocation (Just location) = [("location", location)]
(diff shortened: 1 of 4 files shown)


Yesod/Auth/OAuth2/Github.hs
```diff
@@ -119,3 +119,6 @@ toCreds user userMails token = Creds
 
     maybePublicEmail Nothing = []
     maybePublicEmail (Just e) = [("public_email", e)]
+
+    maybeLocation Nothing = []
+    maybeLocation (Just location) = [("location", location)]
```
pbrisbin: Rule of three says we should think about extracting something now:

```hs
++ maybeKey "name" (githubUserName user)
++ maybeKey "public_email" (githubUserPublicEmail user)
++ maybeKey "location" (githubUserLocation user)

maybeKey :: Text -> Maybe Text -> [(Text, Text)]
maybeKey k (Just x) = [(k, v)]
maybeKey _ Nothing = []

-- Or maybe something fancier, with TupleSections?
maybeKey k = maybeToList . fmap (k,)
```

WDYT? (I'm not in love with the "maybe-key" name, FWIW).

Yesod/Auth/OAuth2/Github.hs
```diff
@@ -47,8 +47,8 @@ instance FromJSON GithubUser where
         <*> o .:? "name"
         <*> o .: "login"
         <*> o .: "avatar_url"
-        <*> o .: "location"
-        <*> o .: "email"
+        <*> o .:? "location"
+        <*> o .:? "email"
```
pbrisbin: Nice catch.

pbrisbin:
I had a random refactoring thought, but this sounds good to me. I don't _think_ there are any compatibility concerns; since the `credsExtra` is just an association list callers should've already been handling a potentially missing key so replacing the error with a missing key sounds like a good bugfix.

Yesod/Auth/OAuth2/Github.hs
adarqui: Nice thnx! Also found two more in Spotify.hs.

adarqui:
A refactor is probably the way to go. Seems like things are handled slightly different in a few of the OAuth2/<Provider>.hs's

I didn't like maybeKey either. But I really couldn't think of anything better.. so I named it "maybeExtra".

I think I like the Spotify.hs style better: the extrasTemplate. Until the code is refactored, that might be the way to go for Github.hs and Google.hs etc, IMHO. I put "maybeExtra" inside of OAuth2.hs, but obviously it shouldn't be there. Those modules should probably share such a function. Maybe a Yesod/Auth/OAuth2/Internal.hs is in order? In there, you (or I) could put the helpers; maybeExtra in this case, or getExtra/mapExtras if you would like to go the Spotify.hs style route.

I'd probably put getExtra and mapExtra (mapMaybe getExtra <extrasTemplate>) in Internal.hs.. include that in the Provider's that use the optional extra credentials, and then change Github.hs/Google.hs to use getExtra/mapExtras & a cleaner extrasTemplate like Spotify.hs uses.. my 2 cents.

What do you think? If you like that, I could make those changes.

Thanks!

pbrisbin:
I actually think this is fine as-is in order to get this fix out without waiting on a broader refactor. My intent was to just extract `maybeExtra` for the 3 uses within `Github.hs` and leave the others duplicated for now, but putting it in the main module is OK too (for now). Cutting some kind of testable `.Internal` module for these things sounds good to me, but can be done after in a new PR.

WDYT @jferris ?

adarqui:
@pbrisbin , Cool! Let me know if you need me to change anything etc.

Thanks!

jferris:
@pbrisbin I think this looks good to merge.

pbrisbin:
Merged as 64b65ca4c6b6c3e6f402b31a682f52124beef4a7, released as 0.1.10. Thanks!

adarqui:
Cool. Thanks!

PR #65, opened 2016-06-24
Sign in with slack

jsteiner:


Yesod/Auth/OAuth2/Slack.hs
```diff
@@ -0,0 +1,113 @@
...
+
+data SlackTeam = SlackTeam
+    { slackTeamId :: Text
+    , slackTeamName :: Text
+    }
+
+instance FromJSON SlackUser where
+    parseJSON = withObject "root" $ \root ->
+        root .: "user" >>= withObject "user"
+            ( \user ->
```
gabebw: Could this be another function, maybe `buildUser`? It could be in a `where`, but the long lambda is hard to read.

gabebw:
Could we add more documentation? e.g.:

* Where to find a Slack token
* An example of it being used in Yesod (`authPlugins [oauth2Slack]` and `authPlugins [oauth2SlackScoped]`, probably?)
* Where can I find the arguments to 
* Can I use `oauth2Slack` and `oauth2SlackScoped` together?

gabebw:
Also, an example of what data might be contained in a returned `SlackUser`.

gabebw:
(This is top-of-mind for me because I just finished integrating with `Yesod.Auth.Dummy`, which is the simplest possible auth plugin, and found it surprisingly difficult. I imagine using a real auth plugin would benefit even more from documentation!)

Yesod/Auth/OAuth2/Slack.hs
```diff
@@ -0,0 +1,112 @@
...
+data SlackTeam = SlackTeam
+    { slackTeamId :: Text
+    , slackTeamName :: Text
+    }
+
+instance FromJSON SlackUser where
+    parseJSON = withObject "root" $ \root -> do
+        user <- root .: "user"
+
+        SlackUser
```
gabebw: ✨

Yesod/Auth/OAuth2/Slack.hs
```diff
@@ -0,0 +1,117 @@
...
+data SlackScope
+    = SlackEmailScope
+    | SlackTeamScope
+    | SlackAvatarScope
+
+data SlackUser = SlackUser
+    { slackUserId :: Text
+    , slackUserName :: Text
+    , slackUserEmail :: Maybe Text
+    , slackUserAvatar :: Maybe Text
```
gabebw: How is this encoded? Is it the avatar URL? Base64-encoded?

jferris:
This looks good to me. @pbrisbin any objections?

README.md
```diff
@@ -22,6 +22,28 @@ clientSecret :: Text
...
+import Yesod.Auth
+import Yesod.Auth.OAuth2.Slack
+
+instance YesodAuth App where
+    -- ...
+
+    authPlugins _ =
+        [oauth2SlackScoped clientId clientSecret slackScopes]
+      where
+        slackScopes = [SlackEmailScope, SlackAvatar, ScopeSlackTeamScope]
```
pbrisbin: Is there a reason two of these end in "Scope" and one doesn't?

Yesod/Auth/OAuth2/Slack.hs
```diff
@@ -0,0 +1,127 @@
...
+    body <- HTTP.responseBody <$> HTTP.httpLbs request manager
+    case eitherDecode body of
+        Left _ -> throwIO $ InvalidProfileResponse "slack" body
+        Right u -> return $ toCreds u token
+
+toCreds :: SlackUser -> AccessToken -> Creds m
+toCreds user token = Creds
+    { credsPlugin = "slack"
+    , credsIdent = slackUserId user
+    , credsExtra = catMaybes
```
pbrisbin: This now the third approach we have to doing this "optional values in the extra association list" thing. I'm fine introducing it as an option to see how it feels, but we should look to centralize them all at some point.

pbrisbin:
I made two small comments but neither are objections. LGTM.

README.md
jferris: Nope. The actual constructor is `SlackAvatarScope`, so this was invalid. Fixed.

Yesod/Auth/OAuth2/Slack.hs
jferris: I've been meaning to look upstream and see if there's a reason we can't expose `SlackUser` to the app. Encoding these as association lists is kind of a pain, it's a pain to decode them again in the app, and it makes it easy to write code that compiles but won't actually work as expected.

PR #68, opened 2016-07-03
Add Bitbucket plugin

pafcu:


pbrisbin:
Sorry for the delayed response here.

I don't have an easy way to test that this works, but I'm inclined to merge it and get it out there for those that want it and let them report back any issues. The code looks like a standard porting of the GH plugin. It's unfortunate that adding these is such a copy-and-paste-like operation, but I don't think that fact should hold up additions like these.

WDYT @jferris @jsteiner ?

pafcu:
It is indeed strongly based on the Github plugin. I have to admit I did not look too closely at the other plugins. Would it make sense to designate one of the plugins (e.g. the Github one) as a "model plugin", making sure that it has extensive documentation (more than what would be necessary for just a standard plugin)? Also, would it make sense to move e.g. the API urls somewhere in the beginning of the file, so that they are all collected in the same place to make it easier to change them when creating a new plugin?

jferris:
I think the GitHub plugin is a good model. I plan on spending some time trying to remove some of the copy-and-paste from the plugins soon. I'm trying to figure out a better way to handle the "extra" data that comes back; I'll open an issue for discussion related to that particular challenge.

pbrisbin:
@jferris would you consider that discussion and resolution blocking this PR? If not, is there anything you'd like to see done here before merging?

jferris:
Nope, this looks good to me. Feel free to merge, or I'll do so on Friday.

jferris:
@pafcu sorry for the delay on this. I merged this into master and will release to Hackage soon.

PR #70, opened 2016-07-06
Rename creds key `email` -> `public_email`

jezen:
We had a name collision when trying to grab the GitHub user's email
address, so if the user did not have a public email address the
authentication would fail with a 500 (InvalidProfileResponse).


pbrisbin:
Fix looks good. Can you back out the version bump? As a matter of process we like to decouple the version change from the actual code changes so we do that ourselves after merging the PR.

jezen:
@pbrisbin Sorry about that; I wasn't familiar with your process. I've backed out that version bump.

pbrisbin:
Merged as 327a85, released as 0.2.1. Thanks!

PR #74, opened 2016-11-12
Relax upper bounds on aeson and http-client

paul-rouse:
I've tested this in a small yesod application, authenticating against Google.  (I compiled it with GHC 8 and used stackage nightly-2016-11-12.)

These bounds are currently blocking this package in stackage nightly, and, when released, I am quite happy to submit a PR to re-enable it.

### DIFF
--- a/yesod-auth-oauth2.cabal
+++ b/yesod-auth-oauth2.cabal
@@ -1,5 +1,5 @@
 name:            yesod-auth-oauth2
-version:         0.2.2
+version:         0.2.3
 license:         BSD3
 license-file:    LICENSE
 author:          Tom Streller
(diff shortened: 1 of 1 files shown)


paul-rouse:
I can't see why the test failed.  The example builds on my system with the "official" stack.yaml, admittedly allowing a newer minor version of GHC, but I doubt if that makes any difference, ie I used: `stack build --flag yesod-auth-oauth2:example --compiler ghc-7.10.3

(That example doesn't build on GHC 8 at the moment because of https://ghc.haskell.org/trac/ghc/ticket/12130.)

paul-rouse:
The test failure isn't anything to with my change, AFAIK.  Is there anything else which needs to be addressed?

paul-rouse:
@jferris - if you're the right person - is there anything I can do to help with pushing this along?  I think the test just needs to be re-run.

sidraval:
Thank you 👍

PR #75, opened 2016-12-15
Add Battle.Net plugin

DuckOfDoom:
Added a minimal Battle.Net plugin for OAuth2 authentication and retrieving user's ID and BattleTag.


sidraval:
Thanks @DuckOfDoom 👍

PR #77, opened 2017-03-24
Just some version bumps for stack 8.5 / ghc 8.x

adarqui:
Just a version bump for http-client and aeson. Worked on my end.

PR #80, opened 2017-05-27
Allow use of any Slack scope

danpalmer:
I'd like to use `yesod-auth-oauth2`'s Slack provider to create a more fully-features Slack application that uses more capabilities of the Slack API. To do this, I need to be able to authenticate with more scopes.

This change removes the `SlackScope` type, and changes the API to take `[Text]` for the scopes (although retains a `defaultSlackIdentityScopes` definition).

I'm relatively new to working on other people's Haskell projects, and to designing Haskell APIs, so please consider this a 'first pass', I'd like any feedback you may have that could help me improve this to the point where it's mergeable.

Note that to get `yesod-auth-oauth2` to compile I did use the changes in #77.

### DIFF
--- a/Yesod/Auth/OAuth2/Slack.hs
+++ b/Yesod/Auth/OAuth2/Slack.hs
@@ -7,7 +7,7 @@
 -- * Returns name, access_token, email, avatar, team_id, and team_name as extras
 --
 module Yesod.Auth.OAuth2.Slack
-    ( SlackScope(..)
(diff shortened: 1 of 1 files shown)


danpalmer:
It turns out the Slack API doesn't actually allow this. You can _either_ sign in, _or_ 'add to Slack', and that these are two different actions, even though you'd often want to combine them.

Closing this PR as it won't work.

PR #81, opened 2017-07-19
Upgrade stack resolver

sidraval:
Currently, `master` doesn't build on Sierra because of GHC 7.10 / stack issues: https://github.com/commercialhaskell/stack/issues/2841

Is there any reason we shouldn't update to the latest resolver?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,8 +2,8 @@ dependencies:
   cache_directories:
     - "~/.stack"
   pre:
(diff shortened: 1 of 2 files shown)


sidraval:
Anyone got any clues about the Circle failure? Running the commands locally builds just fine...

pbrisbin:
The error message indicates to me that it's reading the list of packages as one long string. stack-0.1.6 is really old (I'm on 1.4.0 locally), so I might try updating the CI setup to install new stack?

PR #83, opened 2017-07-28
Update Nylas provider

bts:
Hi there,

Since the last time I used this project, apparently Nylas has changed their API again and the current provider no longer works. The following patch adds `response_type` and `client_id` parameters to fix things up.

Also, because Nylas only allows `scope=email`, I've removed `oauth2NylasScoped`, which wasn't exported from the module anyway.

Thanks,
Brian


pbrisbin:
Thanks!

PR #86, opened 2017-09-25
Update to LTS-9.5 and hoauth2 1.3.0

pbrisbin:
The largest changes were around the hoauth2 interface.

The OAuth2 type replaced all of its ByteString fields with either Text
or URI. This is a huge improvement. The fields that are now Text are the
type we had them in anyway. The fields that are now URI are type safe
and easier to manipulate. For example, we were doing very unsafe query
string manipulations looking for raw ? or & values, but now we can work
with tuples in a well-typed property.

Additionally the AccessToken type was upgraded to OAuth2Token with an
accessToken field, and the simple Either ByteString a results were
replaced by a real OAuth2Error type. This required changes to our
InvalidProfileResponse mechanism to support.

To make working with uri-bytestring more convenient, an Extension
library was added with some useful instances and helper functions. This
library may be upstreamed at some point.


URI/ByteString/Extension.hs
```diff
@@ -0,0 +1,53 @@
...
+        . parseURI strictURIParserOptions
+        . C8.pack
+
+instance IsString (URIRef Relative) where
+    fromString = either (error . show) id
+        . parseRelativeRef strictURIParserOptions
+        . C8.pack
+
+fromText :: Text -> Maybe URI
+fromText = either (const Nothing) Just
```
joshuaclayton: Is there really no function `Either a b -> Maybe b`? Could've sworn this existed.

Yesod/Auth/OAuth2/BattleNet.hs
```diff
@@ -51,30 +51,28 @@ oAuth2BattleNet :: YesodAuth m
                 -> Text -- ^ User region (e.g. "eu", "cn", "us")
                 -> WidgetT m IO () -- ^ Login widget
                 -> AuthPlugin m
-oAuth2BattleNet clientId clientSecret region widget = authOAuth2Widget widget "battle.net" oAuthData (makeCredentials region)
-  where oAuthData = OAuth2 { oauthClientId            = E.encodeUtf8 clientId
-                           , oauthClientSecret        = E.encodeUtf8 clientSecret
-                           , oauthOAuthorizeEndpoint  = E.encodeUtf8 ("https://" <> host <> "/oauth/authorize")
-                           , oauthAccessTokenEndpoint = E.encodeUtf8 ("https://" <> host <> "/oauth/token")
+oAuth2BattleNet clientId clientSecret region widget = authOAuth2Widget widget "battle.net" oAuthData $ makeCredentials region
+  where oAuthData = OAuth2 { oauthClientId            = clientId
```
joshuaclayton: so much nicer!

Yesod/Auth/OAuth2/BattleNet.hs
```diff
@@ -51,30 +51,28 @@ oAuth2BattleNet :: YesodAuth m
...
            { credsPlugin = "battle.net"
            , credsIdent  = T.pack $ show $ userId user
            , credsExtra  = [("battletag", battleTag user)]
            }
-  where host :: URI
-        host = let r = T.toLower region in
-                 case r of
-                   "cn" -> "api.battlenet.com.cn"
-                   _    -> E.encodeUtf8 r <> ".api.battle.net"
+  where host :: Host
```
joshuaclayton: thoughts on pulling this out to its own function?

Yesod/Auth/OAuth2/Bitbucket.hs
```diff
@@ -106,32 +105,34 @@ oauth2BitbucketScoped :: YesodAuth m
...
-    userResult <- authGetJSON manager token "https://api.bitbucket.com/2.0/user"
-    mailResult <- authGetJSON manager token "https://api.bitbucket.com/2.0/user/emails"
+    userResult <- authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user"
+    mailResult <- authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user/emails"
 
     case (userResult, mailResult) of
         (Right user, Right mails) -> return $ toCreds user (bitbucketEmails mails) token
-        (Left err, _) -> throwIO $ InvalidProfileResponse "bitbucket" err
-        (_, Left err) -> throwIO $ InvalidProfileResponse "bitbucket" err
+        (Left err, _) -> throwIO $ invalidProfileResponse "bitbucket" err
```
joshuaclayton: is there anything with `Alternative` that might help here?

Yesod/Auth/OAuth2/Salesforce.hs
```diff
@@ -84,19 +85,21 @@ oauth2SalesforceSandboxScoped scopes clientId clientSecret =
     authOAuth2 svcNameSb oauth fetchSalesforceSandboxUser
   where
     oauth = OAuth2
-        { oauthClientId            = encodeUtf8 clientId
-        , oauthClientSecret        = encodeUtf8 clientSecret
-        , oauthOAuthorizeEndpoint  = encodeUtf8 $ "https://test.salesforce.com/services/oauth2/authorize?scope=" <> T.intercalate " " scopes
+        { oauthClientId            = clientId
+        , oauthClientSecret        = clientSecret
+        , oauthOAuthorizeEndpoint  = "https://test.salesforce.com/services/oauth2/authorize" `withQuery`
+            [ ("scope=", encodeUtf8 $ T.intercalate " " scopes)
```
joshuaclayton: should `"scope="` be `"scope"`?

joshuaclayton: Taking this one step further - thoughts on extracting tuple generation for scope to be shared across each of these?

test/Yesod/Auth/OAuth2Spec.hs
```diff
@@ -4,7 +4,7 @@ module Yesod.Auth.OAuth2Spec
     ) where
 
 import Test.Hspec
-import Yesod.Auth.OAuth2
+--import Yesod.Auth.OAuth2
```
joshuaclayton: Should this be deleted entirely?

URI/ByteString/Extension.hs
pbrisbin: You can do it with Lens or the [either](https://www.stackage.org/haddock/lts-9.6/either-4.4.1.1/Data-Either-Combinators.html) package. I didn't feel it necessary to bring either (heh) of those in. Do you think I should bring in `either` for it?

test/Yesod/Auth/OAuth2Spec.hs
pbrisbin: Yeah, I can actually remove the whole file, now that there are other tests. It was here as a stub about *how* we planned to test even though we had none. It was causing `-Wall` errors to have the unused import, hence the comment. But I'll just nuke it.

URI/ByteString/Extension.hs
jferris: I don't think it's worth the dependency.

joshuaclayton: Agree with @jferris

URI/ByteString/Extension.hs
```diff
@@ -0,0 +1,53 @@
...
+import Data.String (IsString(..))
+import Data.Text (Text)
+import Data.Text.Encoding (decodeUtf8, encodeUtf8)
+import Lens.Micro
+
+import qualified Data.ByteString.Char8 as C8
+
+import URI.ByteString
+
+instance IsString Scheme where
```
jferris: Are the `IsString` instances used outside of the tests?

Yesod/Auth/OAuth2.hs
```diff
@@ -87,15 +98,15 @@ authOAuth2Widget widget name oauth getCreds = AuthPlugin name dispatch login
         tm <- getRouteToParent
         render <- lift getUrlRender
         return oauth
-            { oauthCallback = Just $ encodeUtf8 $ render $ tm url
+            { oauthCallback = Just $ unsafeFromText $ render $ tm url
```
jferris: Does it make sense to wrap this up in a `YesodOAuth2Exception`?

URI/ByteString/Extension.hs
pbrisbin: Yeah, for example: https://github.com/thoughtbot/yesod-auth-oauth2/pull/86/files#diff-c5023724ffce02c103ccf31e67ae3023R57

Yesod/Auth/OAuth2.hs
pbrisbin: Personally, I thought not because it is unexpected enough that the disruption of an `error` would be preferable. In practice, I don't see this ever blowing up because my understating is `render . tm` would either produce a valid URI or blow up itself first.

Yesod/Auth/OAuth2/Salesforce.hs
pbrisbin: Nice catch. I think pulling out a function, while useful for the cases *only* adding `scope`, will make the code where you're adding more than `scope` more cumbersome.

Yesod/Auth/OAuth2/Bitbucket.hs
pbrisbin: Not AFAICT, not without moving up to something like `ExceptT`:

```hs
-- Not tested, but I think this is how you'd do it
result <- runExcepT $ do
    user <- ExceptT $ authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user"
    mails <- ExceptT $ authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user/emails"
    return $ toCreds user (bitbucketEmails mails) token

either (throwIO . invalidProfileResponse "bitbucket") return result
```

With a few ancillary functions, it even makes for a nice `Applicative`:

```hs

throwLeft $ runExcepT $ toCreds' token
    <$> ExceptT (authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user")
    <*> ExceptT (authGetJSON manager (accessToken token) "https://api.bitbucket.com/2.0/user/emails")
  where
    toCreds' t u ms = toCreds u (bitbucketEmails ms) t
    throwLeft = either (throwIO . invalidProfileResponse "bitbucket") return
```

I like this pattern when it's pervasive, but I'm not sure about doing it in a random spot.

Yesod/Auth/OAuth2/Salesforce.hs
pbrisbin: You were right, I'm extracting `scopeParam` :)

PR #91, opened 2017-12-13
changed upper bounds to compile with ghc8.2

### LINKED ISSUE #89: Bump upper bounds to work with stackage-nightly (ghc 8.2)
WARNING: Ignoring out of range dependency (allow-newer enabled): aeson-1.2.3.0. yesod-auth-oauth2 requires: >=0.6 && <1.2
WARNING: Ignoring out of range dependency (allow-newer enabled): hoauth2-1.5.0. yesod-auth-oauth2 requires: >=1.3.0 && <1.4

i had to force "allow-newer". No compile-errors though. Looks like the bump for those two packages would be no problem.
(at least for the current stackage-nightly that will become the next lts-10.x for ghc 8.2)

Drezil:
resolves #89 and #90 ;)

### DIFF
--- a/yesod-auth-oauth2.cabal
+++ b/yesod-auth-oauth2.cabal
@@ -89,6 +89,7 @@ test-suite test
                  , yesod-auth-oauth2
                  , hspec
                  , uri-bytestring
+  other-modules:   URI.ByteString.ExtensionSpec
 
 source-repository head
   type:     git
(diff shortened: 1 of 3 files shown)


yesod-auth-oauth2.cabal
```diff
@@ -89,6 +89,7 @@ test-suite test
                  , yesod-auth-oauth2
                  , hspec
                  , uri-bytestring
+  other-modules:   URI.ByteString.ExtensionSpec
```
pbrisbin: Nice. I wonder if moving to `package.yaml` would address this as well.

pbrisbin:
Committed in #92. Thanks!

Issue #71, opened 2016-07-12
Encoding/Decoding Extras is a Pain


jferris:
We use Yesod's `credsExtra` field on `Creds` to provide user information from remote services. This data is well-formed and includes fields that are likely required for the application, such as the user's email address.

After decoding the remote user into a Haskell value, we de-normalize it into text pairs:

``` haskell
credsExtra :: GithubUser -> [(Text, Text)]
credsExtra user =
        [ ("email", githubUserEmailAddress user)
        , ("login", githubUserLogin user)
        , ("avatar_url", githubUserAvatarUrl user)
        , ("access_token", decodeUtf8 $ accessToken token)
        ]
        ++ maybeExtra "name" (githubUserName user)
        ++ maybeExtra "email" (githubUserPublicEmail user)
        ++ maybeExtra "location" (githubUserLocation user)

maybeExtra :: Text -> Maybe Text -> [(Text, Text)]
maybeExtra k (Just v) = [(k, v)]
maybeExtra _ Nothing  = []
```

We write a function like that for each service.

On the application side, we pairs the text pairs back into a Haskell value:

``` haskell
githubProfile :: [(Text, Text)] -> Either Text Profile
githubProfile extra = Profile
    <$> (lookupExtra "name" extra <|> lookupExtra "login" extra)
    <*> lookupExtra "email" extra

lookupExtra :: Text -> [(Text, b)] -> Either Text b
lookupExtra k extra =
    maybe (Left $ "missing key " ++ k) Right $ lookup k extra
```

The application needs to write a mapping function like this for each service.

There are several issues with this approach:
1. It requires a large amount of copy-and-paste boilerplate for each new service we support.
2. It requires a large amount of boilerplate in the application.
3. Each text key is a value that can be misspelled and break the application. The application will compile successfully but not work.
4. If we change keys in any way on our side, we risk breaking users' applications without breaking the build.
5. The process of discovering which keys are available is poor, because you get an anonymous pair of `Text` values instead of a well-formed Haskell value.

I'd like to do a larger update that improves this experience.

jferris:
The `Creds` type is only parameterized over the Yesod app type, so I don't think we can inject a user type for each service. However, we could expose functions for doing the decoding into well-formed types:

``` haskell
-- In the library
decodeGithubUser :: Creds m -> Either Text GithubUser

-- In the application
githubProfile :: Creds m -> Either Text Profile
githubProfile creds = do
[...]

PR #92, opened 2017-12-29
Provider interface

pbrisbin:
Highlights:

- The "Provider" abstraction, which leads to
- Way tinier definitions, also aided by
- Stop fetching full profiles (#71)

---

In addition to general feedback on this, I plan to do the following before
merging:

- [ ] (Re-)document the heck out of all the modules
- [x] Backfill a bit of a CHANGELOG (so I have a good home to describe *this* major
  change)
- [ ] Go over lost functionality closely, and consider if anything should be
  preserved
- [ ] Take another shot at some kind of actual test suite (Maybe the new break-up
  of functionality can help here)

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -11,7 +11,7 @@ github: thoughtbot/yesod-auth-oauth2.git
 homepage: http://github.com/thoughtbot/yesod-auth-oauth2
 
 dependencies:
-  - base >=4.5 && <5
+  - base >=4.8.0.0 && <5
 
 library:
   source-dirs: src
@@ -21,8 +21,6 @@ library:
     - hoauth2 >=1.3.0 && <1.6
     - http-client >=0.4.0 && <0.6
     - http-conduit >=2.0 && <3.0
-    - http-types >=0.8 && <0.10
-    - lifted-base >=0.2 && <0.4
     - microlens
     - random
     - text >=0.7 && <2.0
--- a/src/URI/ByteString/Extension.hs
+++ b/src/URI/ByteString/Extension.hs
@@ -7,10 +7,9 @@ import Data.String (IsString(..))
 import Data.Text (Text)
 import Data.Text.Encoding (decodeUtf8, encodeUtf8)
 import Lens.Micro
+import URI.ByteString
 
 import qualified Data.ByteString.Char8 as C8
-
-import URI.ByteString
 
 instance IsString Scheme where
     fromString = Scheme . fromString
--- a/src/Yesod/Auth/OAuth2.hs
+++ b/src/Yesod/Auth/OAuth2.hs
@@ -1,167 +1,69 @@
-{-# LANGUAGE CPP #-}
-{-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE QuasiQuotes #-}
-{-# LANGUAGE TupleSections #-}
--- |
---
--- Generic OAuth2 plugin for Yesod
---
--- * See Yesod.Auth.OAuth2.GitHub for example usage.
---
+{-# LANGUAGE RecordWildCards #-}
 module Yesod.Auth.OAuth2
-    ( authOAuth2
-    , authOAuth2Widget
-    , oauth2Url
-    , fromProfileURL
-    , YesodOAuth2Exception(..)
-    , invalidProfileResponse
-    , scopeParam
-    , maybeExtra
-    , module Network.OAuth.OAuth2
-    , module URI.ByteString
-    , module URI.ByteString.Extension
+    ( oauth2Url
+    , authOAuth2
     ) where
 
-#if __GLASGOW_HASKELL__ < 710
-import Control.Applicative ((<$>))
-#endif
-
-import Control.Exception.Lifted
-import Control.Monad (unless)
-import Control.Monad.IO.Class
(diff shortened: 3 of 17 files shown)


pbrisbin:
Needless to say, the diff is rough to review (though the -600 LOC is pretty sweet). I'd recommend viewing the new code and judging things that way.

- https://github.com/thoughtbot/yesod-auth-oauth2/blob/d3fc380f16cc8ee3367d18bb10c23bb729ac1cc2/src/Yesod/Auth/OAuth2.hs
- https://github.com/thoughtbot/yesod-auth-oauth2/blob/d3fc380f16cc8ee3367d18bb10c23bb729ac1cc2/src/Yesod/Auth/OAuth2/Dispatch.hs
https://github.com/thoughtbot/yesod-auth-oauth2/blob/d3fc380f16cc8ee3367d18bb10c23bb729ac1cc2/src/Yesod/Auth/OAuth2/Upcase.hs (very simplest provider)
- https://github.com/thoughtbot/yesod-auth-oauth2/blob/d3fc380f16cc8ee3367d18bb10c23bb729ac1cc2/src/Yesod/Auth/OAuth2/Slack.hs (complex provider)

mwu:
Sorry and I don't mean to butt in but I was using this plugin and stumbled on this pull request. I know it's a WIP, but are there any plans on adding two final decoders for the remaining credsExtra entries, ie `decodeAccessToken :: Creds m -> AccessToken` and `userResponseBody :: Creds m -> Text` or something similar just to take the burden from client code?

Along the same lines, now that clients are more likely to do further fetches in authenticate, should the providers export their names so client code doesn't need to do something like `authenticate c = case credsPlugin c of "github" -> ...; "spotify" -> ...` which feels wrong? I noticed yesod-auth's GoogleEmail2 exports a  [pid](https://hackage.haskell.org/package/yesod-auth-1.4.21/docs/Yesod-Auth-GoogleEmail2.html#v:pid). Seems to be the only plugin that does, though....

Anyway, these changes make using this package a lot easier, so thank you.

pbrisbin:
@mwu thanks a ton for joining in and providing feedback! With such a large change, it's important to me to have some sense that it's a move in the right direction.

> are there any plans on adding two final decoders for the remaining credsExtra entries, ie decodeAccessToken :: Creds m -> AccessToken and userResponseBody :: Creds m -> Text or something similar just to take the burden from client code?

I wasn't planning on that, because the functions would be so simple:

```hs
decodeAccessToken :: Creds m -> Maybe AccessToken
decodeAccessToken = AccessToken <$> lookup "accessToken" . credsExtra

userResponseBody :: Creds m -> Maybe Text
userResponseBody = lookup "userResponseJson" . credsExtra

-- Or perhaps
userResponse :: FromJSON a => Creds m -> Maybe a
userResponse = decode . BL.decodeUtf8 <=< lookup "userResponseJson" . credsExtra
```

If you think it would be useful to provide these, I have no issue adding them.

> now that clients are more likely to do further fetches in authenticate, should the providers export their names

My ultimate goal is that `Provider` is the main API of this package. So in `Foundation`, you might do something like:

```hs
authPlugins = [oauth2Plugin githubProvider cid cs]
```

And you can use `pName githubProvider` anywhere you needed to recover `"github"`. This is the main reason why I don't want you to have to pass the client id/secret just to construct a `Provider` -- `oauth2Plugin` accepts those and puts them together with the provider details at dispatch-time.

This PR is a bit experimental, so as I implement it in stages, there may be times this isn't quite the case yet, but it's where I'm aiming.

> these changes make using this package a lot easier

Are you actually using this branch in a project? I'd be great to know that it's working (we lack pretty much any sort of testing here), and I'm glad to hear the new API is actually easier!

mwu:
Ok let's all avert our eyes a little, because it's pretty rough and I only had a few minutes to play with this last night: https://github.com/mwu/gitit2/commit/8e7518e0ac7b2d847554a7f962906205137cbaa6 But yes, it does work. I haven't yet changed the login widget, but that I can do so is why I was saying this is nicer, among other reasons.

That makes sense now about not needing to repeat the plugin names as literals. For some reason last night it didn't occur to me to define the provider outside the instance definition. The decoder for accessToken I wanted as Creds m -> Text because I'd argue that isn't a real Maybe, but I'm probably just being lazy and trying to throw you a fromJust hot potato. I can't think of anything I'd do with the Nothing case other than throw a generic error, but I grant that I then have the choice as a user what to do with it. It just seems like a lot of ceremony when we're fairly sure it's not going to be a Nothing. The key name's also under your control rather than mine and it's already changed once now, so it's also a bit of a DRY thing 😉

mwu:
Oh and I'm not sure if I detected a hint of worry in your comment, but I'm not using this for work or anything about to go into production. I'm just trying to see how to add oauth to gitit2 which isn't being actively developed. So I'm not rushing to use this, I have whatever time I need to watch this PR and uhhh de-uglify the commit I linked to 😝

pbrisbin:
Thanks so much for providing that context! My plan is to let this PR idle and bring pieces of it slowing in through other, smaller PRs -- rebasing this one along the way until it's reduced to something more manageable.

Adding those decoders would probably be some icing that happens _after_ that long slog of PRs, so we can discuss further the better interface there. I'm not against a `fromJust` when I'm accessing data I've put there myself in the same library, so we'll see.

pbrisbin:
This has served its purpose. I'm going to focus on #100 and follow-up (perhaps) with such a `Provider` abstraction after that's landed.

PR #99, opened 2018-01-26
Test all supported GHCs & nightly

pbrisbin:


jferris:
Cool. I'm assuming the goal is to get the other builds passing before merging?

pbrisbin:
Indeed. The 7.10 job I'll probably disable as TODO, and the 8.0 failure is just a weeder behavior that I don't quite understand (it's installing its own GHC, and not doing it well).

PR #103, opened 2018-03-01
Officially drop 7.10 support

pbrisbin:
In an attempt to get CI to assert it actually worked, I've decided, after many
rounds of extra-deps fighting, that it's either impossible or difficult enough
to give up.

So (among other minor CI tweaks), this set of patches removes the disabled 7.10
CI job and sets the base versions bounds, thus making it official.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,10 +42,6 @@ references:
             fi
 
 jobs:
-  build_7.10.3:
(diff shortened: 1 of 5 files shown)


pbrisbin:
Sigh. I can't reproduce the weeder errors locally, and the example module was most certainly compiled.

PR #114, opened 2018-09-18
Redirect on OAuth2 errors, not permissionDenied

pbrisbin:
And ensure we rescue our InvalidProfile exceptions too.

Overall, this makes the library less likely to cause you 4XX or 5XX
errors due to mis-behaving plugins (and even some expected user actions
such as denying the grant). Instead, we'll always redirect to the login
page with a message set about the issue (for the user) and more details
in the server logs (for you).

### DIFF
--- a/src/Yesod/Auth/OAuth2/Dispatch.hs
+++ b/src/Yesod/Auth/OAuth2/Dispatch.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -19,9 +20,9 @@ import Network.HTTP.Conduit (Manager)
 import Network.OAuth.OAuth2
 import System.Random (newStdGen, randomRs)
 import URI.ByteString.Extension
-import Yesod.Auth
-import Yesod.Auth.OAuth2.ErrorResponse (onErrorResponse)
-import Yesod.Core
+import Yesod.Auth hiding (ServerError)
+import Yesod.Auth.OAuth2.ErrorResponse
+import Yesod.Core hiding (ErrorResponse)
 
 -- | How to take an @'OAuth2Token'@ and retrieve user credentials
 type FetchCreds m = Manager -> OAuth2Token -> IO (Creds m)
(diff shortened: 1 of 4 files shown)


src/Yesod/Auth/OAuth2/Dispatch.hs
```diff
@@ -62,51 +67,63 @@ dispatchCallback name oauth2 getCreds = do
...
+    errLeft :: Show e => IO (Either e a) -> AuthHandler m a
+    errLeft = either (errInvalidOAuth . unknownError . tshow) pure <=< liftIO
 
-    errInvalidOAuth :: (MonadHandler m, MonadLogger m, Show e) => e -> m a
+    errInvalidOAuth :: ErrorResponse -> AuthHandler m a
     errInvalidOAuth err = do
-        $(logError) $ T.pack $ "OAuth2 error: " <> show err
-        permissionDenied "Invalid OAuth2 authentication attempt"
+        $(logError) $ "OAuth2 error (" <> name <> "): " <> tshow err
+        redirectMessage $ "Unable to log in with OAuth2: " <> erUserMessage err
```
eborden: Will `erUserMessage` ever contain sensitive information?

pbrisbin: Nope. That's why it was named with "User" and [documented](https://github.com/thoughtbot/yesod-auth-oauth2/pull/114/files#diff-0f8a730d76fb79da90f465ea7222587eR37) as "for user display".

src/Yesod/Auth/OAuth2/ErrorResponse.hs
```diff
@@ -32,6 +34,25 @@ data ErrorResponse = ErrorResponse
...
+    AccessDenied -> "Access denied"
+    UnsupportedResponseType -> "Unsupported response type"
+    InvalidScope -> "Invalid scope"
+    ServerError -> "Server error"
+    TemporarilyUnavailable -> "Temporarily unavailable"
+    Unknown _ -> "Unknown error"
+
+unknownError :: Text -> ErrorResponse
+unknownError x = ErrorResponse
+    { erName = Unknown x
```
pbrisbin: As an alternative, I could dump this `x` into `Description` instead, and leave the `Name` as a value-less `Unknown`. Then `erUserMessage` may be unneccessary as a simple `show`-like on `ErrorName` couldn't contain any details. :man_shrugging:

PR #118, opened 2018-10-31
Add AzureAD provider

cbeav:
`hoauth2-1.8.1` incorporated [my PR](https://github.com/freizl/hoauth2/pull/97) for more flexible `expiresIn` parsing on `OAuth2Token`. With this, it's now super straightforward to add support for AzureAD as a provider.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [*Unreleased*](https://github.com/thoughtbot/yesod-auth-oauth2/compare/v0.6.1.0...master)
 
 - Test with GHC 8.6.3, and not 8.2
+- Added AzureAD provider
+- COMPATIBILITY: Use `hoauth2-1.8.1`
 
 ## [v0.6.1.0](https://github.com/thoughtbot/yesod-auth-oauth2/compare/v0.6.0.0...v0.6.1.0)
 
--- a/src/Yesod/Auth/OAuth2/AzureAD.hs
+++ b/src/Yesod/Auth/OAuth2/AzureAD.hs
@@ -1,0 +1,53 @@
+{-# LANGUAGE OverloadedStrings #-}
+-- |
+--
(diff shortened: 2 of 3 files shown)


cbeav:
Is there a reason why `hoauth2` version is capped to `<1.8` in `package.yaml`?

pbrisbin:
I just put upper bounds as a matter of course (this is a practice I'm considering dropping, but it's here for now). So feel free to bump iti in this PR.

cbeav:
Agh. So `8.2.2` is failing to build because it can't find `(<>)`. I'm only loosely familiar with [the GHC changes](https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid) to contain `Semigroup` within `Monoid` and promote `semigroups` into `base`, and I don't know what versions all these changes landed in from memory, but does it seem like this is indeed the problem? If so, how long are you hoping to continue supporting GHC 8.2.2 and should I pursue upstream changes to avoid `(<>)`?

pbrisbin:
My general stance is to support 1 major GHC version back from current, where current is defined as what's in the latest Stackage LTS. So, at this time, that's still 8.4 and 8.2. However, 8.6 is already released and will be in the next LTS very soon I'm sure, so given that the `hoauth2` we require has (perhaps unknowingly) dropped 8.2 support, I think we can too.

The only other option (I think) is to report an Issue (or open a PR) on `hoauth2` making it compile on 8.2 as well (perhaps with `CPP`). Is that something you'd want to do?

cbeav:
I think I'll just park on this until 8.6 LTS lands. Happy to update the Circle CI builds too as part of that. You okay with this staying open til then?

cbeav:
Woooo, LTH Haskell 13.0 shipping with GHC 8.6.3, a true Christmas miracle!

pbrisbin:
Is this ready to merge?

cbeav:
Good to go!

PR #119, opened 2018-11-12
Fix a link in package.yaml

chromezh:
The `.git` is redundant because on hackage the link of Bug tracker would be `https://github.com/thoughtbot/yesod-auth-oauth2.git/issues` and it is not a valid link.


pbrisbin:
Thanks!

PR #130, opened 2020-07-08
Add WordPress.com as an auth provider

nbloomf:
The immediate purpose of this PR is to aid in testing #128, as the WordPress.com oauth2 implementation requires form parameter style client credentials when fetching the token. That PR has a link to a live test implementation.

Documentation for this integration is at https://developer.wordpress.com/docs/wpcc/


pbrisbin:
Sorry I lost track of this! I'll merge it this week, once I work out the CI issues in `pb/next`.

PR #142, opened 2021-02-26
Consolidate all errors, use onErrorHtml

pbrisbin:
Prior to this commit, some errors would be thrown (missing parameter, invalid
state, incorrect approot) while others would be handled via the
set-message-redirect approach (handshake failure, fetch-token failure, etc).

This commit consolidates all of these cases into a single `DispatchError` type,
and then uses `MonadError` (concretely `ExceptT`) to capture them all and handle
them in one place ourselves.

It then updates that handling to:

- Use `onErrorHtml`

  `onErrorHtml` will, by default, set-message-redirect. That make this behavior
  neutral for users running defaults. For users that have customized this, it
  will be an improvement that all our error cases now respect it.

- Provided a JSON representation of errors
- Attach a random correlation identifier

The last two were just nice-to-haves that were cheap to add once the code was in
this state.

Note that the use of `MonadError` requires a potentially "bad" orphan
`MonadUnliftIO` instance for `ExceptT`, but I'd like to see that instance become
a reality and think it needs some real-world experimentation to get there, so
here I am.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -35,11 +35,13 @@ library:
     - http-types >=0.8 && <0.13
     - memory
     - microlens
+    - mtl
     - safe-exceptions
     - text >=0.7 && <2.0
     - uri-bytestring
     - yesod-auth >=1.6.0 && <1.7
     - yesod-core >=1.6.0 && <1.7
+    - unliftio
 
 executables:
   yesod-auth-oauth2-example:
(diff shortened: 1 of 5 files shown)


pbrisbin:
First commit is the thing, with a minimized diff. Subsequent diffs are refactor.

halogenandtoast (approved):
I wish I had an opinion on the orphan instance. It seems fine enough to me, but my experience with unliftio is rather limited. Overall the direction looks good and the new features seem great.

PR #149, opened 2021-03-07
Expose custom widgets for google oauth

jmorag:
I needed this for google specifically, but I'm happy to apply similar changes to the other providers that don't accept custom widget arguments (basically all the non-gaming ones).


pbrisbin:
Nice, thanks!

We do something similar for [`EveOnline`](https://github.com/freckle/yesod-auth-oauth2/blob/main/src/Yesod/Auth/OAuth2/EveOnline.hs), but it offers a set of alternatives with choice by enum, rather than going completely custom. There may be a broader way to solve configurability and normalize with that.

For now, I suggest we keeping adding this piece-meal as needed and maybe come back with an overall solution after a few more have cropped up.

pbrisbin (approved):

PR #150, opened 2021-04-09
Expose `onDispatchError` and generic error message

mjgpy3:
For pass through logging.

I'd also like to enable our release github action. Will the pattern we use in bcp suffice (since this has multiple exes)?


pbrisbin (approved):
I don't think you'll need anything special for release. Only multi-package (not multi-exe) makes it complicated.

PR #152, opened 2021-04-09
Add release action

mjgpy3:
Opened for visibility, will be merging on green.


mjgpy3:
Erm, maybe before green since `ci.yml` shouldn't really care about this.

PR #157, opened 2022-02-28
Support hoauth2-2.2/3

pbrisbin:
This required a lot of CPP refactoring and extension. I plan to shift our lower
bound and target only the newer `hoauth2` soon, but I'd like to get out a
compatible version first, which this aims to do.

The comments in `Compat.hs` try to explain the gymnastics we have to endure to
get there. I'm sorry, it's not ideal.


mjgpy3 (approved):
I think this makes sense. I assume existing `stack-yaml`s in the CI matrix compile against hoauth<2 and that confirms the safety of this change?

pbrisbin:
Yup, `stack.yaml` is v18 and has hoauth2-1.16. The older resolvers are even older, but I'm not sure what exactly.

https://www.stackage.org/package/hoauth2/snapshots

PR #159, opened 2022-03-18
Add twitch.tv plugin

chiroptical:
I recently put together a Twitch plugin. Let me know if there is anything I can do to test. My guess is that you will need to create credentials at https://dev.twitch.tv, plug them into the Heroku instance, and then it should work.


yesod-auth-oauth2.cabal
```diff
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: d3ba666a53fb63a2f059501ac04eca709a905d0dd1f806ea5a7cd64e06ef890d
+-- hash: 233909874fdbdbd71fa70c49f5a4223b4150b85d9415dbbed7fde2fff9e5ebcf
```
chiroptical, line 7: I feel like this is probably a bad change?

pbrisbin: No it's good. It just happens any time this file's contents change, and you added a module.

src/Yesod/Auth/OAuth2/Twitch.hs
```diff
@@ -0,0 +1,56 @@
...
+ where
+  oauth2 = OAuth2
+    { oauth2ClientId          = clientId
+    , oauth2ClientSecret      = Just clientSecret
+    , oauth2AuthorizeEndpoint = "https://id.twitch.tv/oauth2/authorize"
+                                  `withQuery` [scopeParam " " scopes]
+    , oauth2TokenEndpoint     = "https://id.twitch.tv/oauth2/token"
+                                  `withQuery` [ ("client_id", T.encodeUtf8 clientId)
+                                              , ("client_secret", T.encodeUtf8 clientSecret)
+                                              ]
```
chiroptical, line 54: These both must be passed to `/oauth2/token`, see https://dev.twitch.tv/docs/authentication/getting-tokens-oauth#oauth-authorization-code-flow step 3.

chiroptical:
I didn't see a `format` or `weeder` make target, might be useful to provide or give instructions.

pbrisbin:
It's OK. I would just merge the Restyled PR if it mattered -- or there are instructions in its PR description if you'd like to incorporate its changes locally. I handle auto-formatting locally through editor integrations, hence no `make` target.

In this case, this (older) project is a little too mis-formatted and I need to make a one-time formatting pass anyway, so I'm inclined to ignore it for now.

If you're happy with this PR, I'll merge on green CI.

pbrisbin (approved):

chiroptical:
Nice. I am good with this. I want to open up another PR to explain the callback thing but I think that happens upstream.

PR #161, opened 2022-03-25
Add nightly on CI

pbrisbin:
LTS-19 and GHC-9.2 is coming. Adding this to ensure we can compile,
which seems to be the case.

We are only disabled in Stackage because hoauth2 is not ready, for which
I've opened: https://github.com/freizl/hoauth2/issues/142


stack-nightly.yaml
```diff
@@ -0,0 +1,11 @@
+resolver: nightly-2022-03-25
```
mjgpy3, line 1: So does this get overwritten by `--resolver nightly`?

mjgpy3 (approved):

stack-nightly.yaml
pbrisbin: Yeah, unfortunately `resolver: nightly` in a `stack.yaml` doesn't work. So this basically serves as "when was the last time we worked on this?" and you must use `--resolver` to test actual now-nightly.

PR #165, opened 2023-01-30
Fix hoauth2 compat for 2.7.0

mjgpy3:
Fix compatibility.


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## [_Unreleased_](https://github.com/thoughtbot/yesod-auth-oauth2/compare/v0.7.0.2...main)
+## [_Unreleased_](https://github.com/thoughtbot/yesod-auth-oauth2/compare/v0.7.0.3...main)
```
pbrisbin, line 1: I'd either remove this (and version bump in its own PR), or include the `package.yaml` / `.cabal` change to `version` here too.

src/Network/OAuth/OAuth2/Compat.hs
```diff
@@ -48,14 +68,14 @@ fetchAccessToken
     :: Manager
     -> OAuth2
     -> ExchangeToken
-    -> IO (OAuth2Result Errors OAuth2Token)
+    -> IO (OAuth2Result OAuth2Token)
```
pbrisbin, line 71: Don't changes in our own signatures like this mean we have to be a major version bump (`0.8.0.0`)?

src/Network/OAuth/OAuth2/Compat.hs
```diff
@@ -39,7 +53,13 @@ data OAuth2 = OAuth2
...
     }
 
-type OAuth2Result err a = Either (OAuth2Error err) a
+#if MIN_VERSION_hoauth2(2,7,0)
+type Error = TokenRequestError
+#else
+type Error = OAuth2Error Errors
+#endif
+
+type OAuth2Result a = Either Error a
```
pbrisbin, line 62: Why do we have to export `Error` if we're speaking in terms of this, which kind of hides it?

stack-hoauth2-2.6.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+extra-deps:
+  - hoauth2-2.6.0
+
+# hoauth2 needs upper-bounds relaxed for
+#
+# - memory-0.18.0
+# - text-2.0.1
+#
+allow-newer: true
```
pbrisbin, line 11: There's a key in newer Stack where you can say which dependencies in specific you are allowing, rather than just ignoring all bounds ever. I'd for sure use that here now that it's available.

pbrisbin (changes requested):
Thankfully, this project has exhaustive CI across all the dependency versions -- so as long as it's all green, then LGTM. I'm just blocking on the version-bump stuff.

src/Network/OAuth/OAuth2/Compat.hs
mjgpy3: Yep, you're right. Will change.

src/Network/OAuth/OAuth2/Compat.hs
mjgpy3: `Error` gets [rewrapped](https://github.com/freckle/yesod-auth-oauth2/blob/d8011561b8159d482b3aa300cfed5d20fc1353b7/src/Yesod/Auth/OAuth2/DispatchError.hs#L34) and potentially "re-raised" when we [try to get a token](https://github.com/freckle/yesod-auth-oauth2/blob/36805f0580bedec39a69c254456df14474f3bec1/src/Yesod/Auth/OAuth2/Dispatch.hs#L89).

mjgpy3: Would you recommend keeping the type parameter so it's less hidden?

src/Network/OAuth/OAuth2/Compat.hs
pbrisbin: I guess also, is there any way to avoid it?

src/Network/OAuth/OAuth2/Compat.hs
pbrisbin: > Would you recommend keeping the type parameter so it's less hidden?

I don't mind so much about hidden, but I think keeping the type parameter allows us to avoid ourselves being a major version bump. If so, I'd recommend it for that reason.

src/Network/OAuth/OAuth2/Compat.hs
mjgpy3: It's tricky because the <2.7.0 error type is gone... As linked elsewhere, we re-throw that error.

So from a compat perspective I'm not really sure how to get around that (albeit implicit in an error's type) interface change.

pbrisbin: Do you truly mean "throw", or return as a potential `Left`?

pbrisbin: If we introduce our own `Error` that works the same to any possible use case (i.e. a pattern match), I'd consider that backwards compatible. It doesn't matter if we supply it now that the library does not.

It's a separately question how much we want to bend over backwards to continue to avoid v0.8 (or even v1.0). If we think we're at that point to bump we should consider moving our lower bound on hoauth2 up and ditching _all_ the CPP. Yeah?

mjgpy3: So we do a true throw but it looks like we catch it before it gets to possible downstream consumers.

It looks like this handling results in a 
 - log message (this will change but it's just a log, right? Right? A log format shouldn't change our interface, right?)
 - HTML page containing `dispatchErrorMessage` which doesn't care about the underlying, expecific error

So, I'll take a stab at making this backwards compatible.

mjgpy3: Looks like `OAuth2Error Errors` leaks to higher stuff than just the handler I'm looking at :( (e.g. stuff in our Prelude)... Not sure how promising this is feeling... I'lll see if there's something obvious but I'm leaning more towards the "ditch the CPP" approach.

pbrisbin: I'm curious why it bubbling up higher making it feel any less promising. Isn't the question just if we can preserve all the same interfaces or not? How many things rely on those interfaces doesn't really change the work to be done right here, does it?

mjgpy3:
@pbrisbin this PR is now up-to-date with our offline discussion.

For those watching at home, we're going with a minor version update because we do not believe this will break downstream consumers.

src/Network/OAuth/OAuth2/Compat.hs
mjgpy3: Talked it through offline. These concerns should all be internal to the package and should not affect downstream consumers.

src/Network/OAuth/OAuth2/Compat.hs
mjgpy3: Another essentially internal type.

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
...
+
+## [v0.7.0.3](https://github.com/thoughtbot/yesod-auth-oauth2/compare/v0.7.0.2...v0.7.0.3)
+
+- Support `hoauth-2.7`, this change is breaking if something other than
+  than `fetchAccessToken` or `fetchAccessToken2` is used build the plugin (which
+  we don't believe to be the case)
```
pbrisbin, line 7:
```suggestion
- Support `hoauth-2.7`. This change is only breaking in the unlikely case of users
  using something other than `fetchAccessToken` or `fetchAccessToken2`
```

src/Network/OAuth/OAuth2/Compat.hs
```diff
@@ -15,15 +16,28 @@ module Network.OAuth.OAuth2.Compat
...
     , OAuth2Token(..)
     , RefreshToken(..)
     )
+#endif
 import qualified Network.OAuth.OAuth2 as OAuth2
+#if MIN_VERSION_hoauth2(2,7,0)
+import Network.OAuth.OAuth2.TokenRequest (TokenRequestError)
+#else
 import Network.OAuth.OAuth2.TokenRequest (Errors)
+#endif
```
pbrisbin, line 40: Can this be consolidated with the same condition at L19/26?

src/Network/OAuth/OAuth2/Compat.hs
```diff
@@ -48,14 +68,14 @@ fetchAccessToken
     :: Manager
     -> OAuth2
     -> ExchangeToken
-    -> IO (OAuth2Result Errors OAuth2Token)
+    -> IO (OAuth2Result Error OAuth2Token)
```
pbrisbin, line 71: Why can't we avoid this diff?

```hs

#if MIN_VERSION_hoauth2(2,7,0)
type Errors = TokenRequestError
#else
type Errors = OAuth2Error OAuth2.Errors
#endif

type OAuth2Result err a = Either err a

-- ...

   -> IO (OAuth2Result Errors OAuth2Token)
```

Sorry if this is re-hashing trodden ground. I'm not blocking on it.

pbrisbin (approved):

PR #169, opened 2023-04-05
Add AzureADv2 plugin

pbrisbin:
This is the same as the `AzureAD` plugin except:

1. It uses tenant-specific `microsoftonline.com` v2 OAuth2 endpoints
   (hence the name), which means accepting a new Tenant Id argument
2. It uses a space instead of `,` as the scopes separator

Users of multi-tenant apps can provide a Tenant Id of `"common"`. I'm
also not certain if the space-vs-comma scopes separator represents a bug
in the `AzureAD` plugin, or just a difference in the actual v2 APIs.

This inherits the behavior of using email address as the `credsIdent`
even though there is definitely an `id` field in the User Response. I'm not
sure if there are trade-offs one way or another. Using `id` could mean
transparently handling Azure users changing their email, but I suspect
your identity is implicitly tied to email within Azure anyway, so that
would not be a case we'll ever see.

In the future, we can deprecate the `AzureAD` plugin and suggest users
migrate to this one.


src/Yesod/Auth/OAuth2/AzureADv2.hs
```diff
@@ -0,0 +1,82 @@
...
+defaultScopes :: [Text]
+defaultScopes = ["openid", "profile"]
+
+oauth2AzureADv2
+  :: YesodAuth m
+  => Text
+    -- ^ Tenant Id
+    --
+    -- If using a multi-tenant App, @common@ can be given here.
+    --
```
pbrisbin, line 37:
```suggestion
  -- ^ Tenant Id
  --
  -- If using a multi-tenant App, @common@ can be given here.
  --
```

Brittany fixed the signature, but not the comment.

src/Yesod/Auth/OAuth2/AzureADv2.hs
```diff
@@ -0,0 +1,82 @@
...
+oauth2AzureADv2 = oauth2AzureADv2Scoped defaultScopes
+
+oauth2AzureADv2Scoped
+  :: YesodAuth m
+  => [Text] -- ^ Scopes
+  -> Text
+    -- ^ Tenant Id
+    --
+    -- If using a multi-tenant App, @common@ can be given here.
+    --
```
pbrisbin, line 50:
```suggestion
  -- ^ Tenant Id
  --
  -- If using a multi-tenant App, @common@ can be given here.
  --
```

joelmccracken (approved):

PR #171, opened 2023-04-05
pb/prefactor

pbrisbin:
- Add brittany.yaml
- Reformat everything with Brittany
- Reformat everything with Stylish Haskell
- Add shellcheck pragma to .env.example
- White-space
- Document Example in README


pbrisbin:
Merging without review as it's low-impact, behavior-neutral, and blocking my actual PR.

PR #173, opened 2023-04-06
Refactor stack matrix

pbrisbin:
- Use our conventional resolvers by GHC
- Use hoauth2-2.8.0 in nightly, instead of allow-newer-deps
- Document matrix in CI workflow source


pbrisbin:
I'll have to adjust Required Statuses when this is lands. This PR itself may not be Green in the meantime.

stackptr (approved):
YAML changes look coherent but I don't have enough context to validate various version changes

PR #174, opened 2023-06-30
Add Okta plugin and relative URL support

cptrodolfox:
Adds support for: 

- Okta with prompt requirement to authorize redirect
- Relative application roots using `oauth2AppRoot`


src/Network/OAuth/OAuth2/Compat.hs
```diff
@@ -44,6 +44,7 @@ data OAuth2 = OAuth2
     , oauth2AuthorizeEndpoint :: URIRef Absolute
     , oauth2TokenEndpoint :: URIRef Absolute
     , oauth2RedirectUri :: Maybe (URIRef Absolute)
+    , oauth2AppRoot :: Maybe Text
```
pbrisbin, line 47: I don't think we want to expand this type with fields not present in [`hoauth2`s type](https://hackage.haskell.org/package/hoauth2-2.8.1/docs/Network-OAuth-OAuth2.html#t:OAuth2). Some do we hope to drop compatibility with older versions and just use it directly. A change like this would block that.

src/Yesod/Auth/OAuth2/Dispatch.hs
```diff
@@ -100,7 +100,11 @@ withCallbackAndState
   -> Text
   -> m OAuth2
 withCallbackAndState name oauth2 csrf = do
-  uri <- ($ PluginR name ["callback"]) <$> getParentUrlRender
+  pluginURI <- ($ PluginR name ["callback"]) <$> getParentUrlRender
+  let
+    uri = case oauth2AppRoot oauth2 of
+      Just root -> root <> pluginURI
+      Nothing -> pluginURI
```
pbrisbin, line 107: Can you explain why you found this necessary?

If I understand it correctly, `getParentUrlRender` will respect [`Yesod(approot)`](https://hackage.haskell.org/package/yesod-core-1.6.24.4/docs/Yesod-Core.html#v:approot). Are you finding that you need to do to support a root here, but cannot use an overall [`Approot`](https://hackage.haskell.org/package/yesod-core-1.6.24.4/docs/Yesod-Core.html#t:Approot) value to accomplish it?

pbrisbin (commented):
Thanks! The Plugin looks good, but I think we have to figure out the approot thing differently. Let me know your thoughts.

Also, once #176 lands, I'll ask that you rebase and reformat with Fourmolu but that should be an easy thing.

src/Network/OAuth/OAuth2/Compat.hs
cptrodolfox: I am with you regarding this change, this was a quick hack to be able to use relative path as the app root.

src/Yesod/Auth/OAuth2/Dispatch.hs
cptrodolfox: The problem was that in the project I integrated this library with, the `approot` was set to `ApprootRelative` . Thus, the value generated by `getParentUrlRender` would have been a relative path, e.g. `/auth/plugin/okta/callback`.

cptrodolfox: @pbrisbin Do you have any suggestions on how to better handle this case?

pbrisbin: There is an existing limitation that a project can't use a relative Approot and use this library -- therefore, this plugin can rely on it. I should probably dig up what that limitation was and properly document this :thinking:

pbrisbin: Oh it's "documented" here: https://github.com/freckle/yesod-auth-oauth2/blob/main/example/Main.hs#L56-L58

cptrodolfox: @pbrisbin What do you think about support for Yesod projects with relative roots?

pbrisbin: Feel free to work on it, but we've been around a long time with this restriction in place and it's been OK. As far as I know OAuth2 (the protocol) needs to know domain for many reasons. So for use to lift our restriction on apps needing a static approot to use this library, we'd have to resort to some sort of "guess approot" function, which I think ultimately falls back on the `Host` header, which can be spoofed, which feels risky -- so I would need some convincing that it's safe to go that direction.

cptrodolfox: @pbrisbin Could we change the behavior of `oauth2RedirectUri`? So, if the value is already set and valid we do no change it to the built `pluginURI` in the function `withCallbackAndState`.

cptrodolfox: That way we can have support for relative appRoots without this hack.

pbrisbin: > Could we change the behavior of oauth2RedirectUri?

Hmm, I don't think so. That behavior (that field) comes directly from `hoauth2`:



And its type (`URIRef Absolute`) is making it very clear, IMO, that it simply must be absolute. Therefore, to use OAuth2, you must use an absolute appRoot.

cptrodolfox: I have created this PR to illustrate the change:
https://github.com/freckle/yesod-auth-oauth2/pull/178

pbrisbin:
Can we merge this PR such that it just relies on the existing limitation that the app is using a static approot and then tackle relative approots separately?

cptrodolfox:
> Can we merge this PR such that it just relies on the existing limitation that the app is using a static approot and then tackle relative approots separately?

Yes, I will remove the change for relative `approots` on this PR. We can continue the discussion regarding support of relative approots in the other PR https://github.com/freckle/yesod-auth-oauth2/pull/178.

pbrisbin:
:wave: what's the status on this PR? I think it just needs a rebase now that #178 landed right?

pbrisbin:
@cptrodolfox bump

pbrisbin:
@cptrodolfox bump

I'll probably close this as stale if there's no response this week.

PR #178, opened 2023-10-31
Add support for Relative Approots non intrusive

cptrodolfox:
By using an alternative with oauth2RedirectUri users of the library can set up their own redirect URIs.

### DIFF
--- a/src/Yesod/Auth/OAuth2/Dispatch.hs
+++ b/src/Yesod/Auth/OAuth2/Dispatch.hs
@@ -12,8 +12,8 @@ module Yesod.Auth.OAuth2.Dispatch
   , dispatchAuthRequest
   ) where
 
-import Control.Monad (unless)
 import Control.Applicative ((<|>))
+import Control.Monad (unless)
 import Control.Monad.Except (MonadError (..))
 import Data.Text (Text)
 import qualified Data.Text as T
--- a/src/Yesod/Auth/OAuth2/Dispatch.hs
+++ b/src/Yesod/Auth/OAuth2/Dispatch.hs
@@ -105,7 +105,7 @@ withCallbackAndState
   -> Text
   -> m OAuth2
 withCallbackAndState name oauth2 csrf = do
-  callback <- maybe uriFromPlugin pure $ oauth2RedirectUri oauth2
+  callback <- maybe defaultCallback pure $ oauth2RedirectUri oauth2
   pure
     oauth2
       { oauth2RedirectUri = Just callback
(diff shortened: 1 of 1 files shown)


pbrisbin (approved):
This almost feels too easy!

Let me make sure I understand...

Previously,

Plugins would create an `OAuth2` with `oauth2RedirectUri` set to `Nothing`, and this function would then build a `callback` using `PluginR` and set it to that. This relies on `getParentUrlRender` which relies on `approot` being non-relative.

Now,

Plugins may set an `oauth2RedirectUri` ahead of time and it will be respected.

This allows an app that uses a relative approot to function, provided they only use _plugins_ that are handling `oauth2RedirectUri` themselves, and ensuring it's absolute through whatever plugin-specific variables they may accept.

This seems kind of niche to me, but it's very non-intrusive and totally opt-in, so sounds great!

src/Yesod/Auth/OAuth2/Dispatch.hs
```diff
@@ -109,7 +110,7 @@ withCallbackAndState name oauth2 csrf = do
   callback <- maybe (throwError $ InvalidCallbackUri uri) pure $ fromText uri
```
pbrisbin, line 110: Should be move the defaulting up here such that `oauth2RedirectUri` being `Just` will avoid the possibility of `throwError` altogether?

pbrisbin: For example,

```hs
withCallbackAndState name oauth2 csrf = do
  callback <- maybe getDefaultCallback pure oauth2RedirectUri oauth2

  pure
    oauth2
      { oauth2RedirectUri = Just callback

getDefaultCallback = do
  uri <- ($ PluginR name ["callback"]) <$> getParentUrlRender
  maybe (throwError $ InvalidCallbackUri uri) pure $ fromText uri
```

cptrodolfox:
@pbrisbin I have done some code changes to avoid calling the `getParentUrlRender` if the `oauth2RedirectUri` is set.

cptrodolfox:
> LGTM. I will merge/release once I see it's Green.

Thanks for the feedback.

pbrisbin:
Released as v0.7.1.3.

DavidMazarro:
@cptrodolfox Thanks for the contribution! Remember to add the following line in the PR's description so that the contribution can be considered valid for Stack Builders' Hacktoberfest 2023 event:

> As part of [Stack Builders Inc.](https://www.stackbuilders.com/) Hacktoberfest 2023 event.

PR #183, opened 2024-05-24
orcid auth plugin

c-ding-math:


src/Yesod/Auth/OAuth2/ORCID.hs
```diff
@@ -0,0 +1,50 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Yesod.Auth.OAuth2.Orcid
```
pbrisbin, line 3: I see the module name is `Orcid` but the file is `ORCID.hs`. I think these should match. I worry about problems on MacOS (case-insensitive file system). I would probably use `ORCID` consistently.

pbrisbin (commented):
This is looking good, can you apply the Restyled fixes? There are instructions in the build log if you click the details link on the failing PR status.

src/Yesod/Auth/OAuth2/ORCID.hs
c-ding-math: I've corrected the module name.

src/Yesod/Auth/OAuth2/ORCID.hs
```diff
@@ -0,0 +1,50 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Yesod.Auth.OAuth2.ORCID
+  ( oauth2Orcid
```
pbrisbin, line 4: Thanks for correcting the module name. I think I'd like the capitalization to be consistent everywhere.

```suggestion
  ( oauth2ORCID
```

pbrisbin (changes requested):

src/Yesod/Auth/OAuth2/ORCID.hs
c-ding-math: Done

pbrisbin:
Re-opened in #186, will go out with v0.7.3.0.

PR #184, opened 2024-07-08
Replace cryptonite with crypton

pbrisbin:
https://github.com/commercialhaskell/stackage/issues/7474#issuecomment-2208142024


pbrisbin:
I decided to drop lts-14 / ghc-8.6 from CI. I got the error `crypton is a library package but provides no library`, which I didn't want to investigate further. It may still build, we we'll no longer claim it as supported (e.g. tested) and we'll push our lower bounds up to match lts-16 now instead.

mjgpy3 (approved):

